PR #3, opened 2020-04-29
Rename ioctl prefix and communication structures

aserdean:
This PR includes two changes.

One change renames the structure used by the kernel/userspace communication:
USER_IN -> CONNECTION_INFO
LIST_ENTRY_OUT -> DISK_INFO
GET_LIST_OUT -> DISK_INFO_LIST

The second change drops the `VM` (virtual media) part from the prefix `IOCTL_WNBDVM`.


driver/userspace_shared.h
```diff
@@ -40,25 +40,25 @@ typedef struct _USER_IN {
     INT             Pid;
     UINT64          DiskSize;
     UINT16          BlockSize;
-} USER_IN, * PUSER_IN;
+} CONNECTION_INFO, * PCONNECTION_INFO;
 
 typedef struct _USER_COMMAND {
```
petrutlucian94, line 45: Do we really need a structure? Can we do a typedef? Also, a better name might be "WNBD_COMMAND"

aserdean: As discussed offline:
`
it' usually just better to do sizeof struct instead of something that can hold the ioctl code and also it's easier if we want to add something more instead of just the ioctl code i.e. feed other flags like operational mode.
`

I will change the naming in a different PR since I want to push more changes beside this one.

Closing this PR and opening a new one.

aserdean:
Superseded-by: https://github.com/cloudbase/wnbd/pull/4

PR #4, opened 2020-04-29
Userspace and kernelspace polishing

aserdean:
This PR contains the following changes:

- Renaming changes:
USER_IN -> CONNECTION_INFO
LIST_ENTRY_OUT -> DISK_INFO
GET_LIST_OUT -> DISK_INFO_LIST
USER_COMMAND -> WNBD_COMMAND
Drops the `VM` (virtual media) part from the prefix `IOCTL_WNBDVM`.

- Styling:
`wnbd-client` output was off by one tab.
Do not print the Status instead return it.
Drop the printing of the input parameters when using `map` command.
Drop the prefix when converting the get last error code to string.

- Functional
Passing a `NULL` buffer to the listing IOCTL now returns the needed buffer size. If the buffer size is not enough the user will get an `STATUS_BUFFER_OVERFLOW` error.
Use the driver defined number of targets instead of the `<storport.h>` one.


petrutlucian94:
Thanks for addressing those issues!

PR #5, opened 2020-05-29
Fixes needed for certification tests

aserdean:


driver/wnbd.inf
```diff
@@ -50,8 +51,8 @@ LoadOrderGroup = SCSI Miniport
 AddReg         = pnpsafe_isa_addreg
 
 [pnpsafe_isa_addreg]
-HKR,,"BreakOnEntry", %REG_DWORD%,  0x00000000
-HKR,"Parameters","BusType",%REG_DWORD%,0x0000000E
+HKR, "Parameters\PnpInterface", "5", %REG_DWORD%, 0x00000001
```
petrutlucian94, line 54: Ok, so we're using "5" because it's a PCI bus.

driver/scsi_driver_extensions.c
```diff
@@ -277,6 +278,142 @@ WnbdHwResetBus(PVOID DeviceExtension,
...
+        goto exit;
+    }
+
+    RtlZeroMemory((PCHAR)Info, Request->DataBufferLength);
+
+    /* Just fill data needed for WLK Firmware update test */
+    Info->Version = STORAGE_FIRMWARE_INFO_STRUCTURE_VERSION_V2;
+    Info->Size = sizeof(STORAGE_FIRMWARE_INFO_V2);
+    Info->UpgradeSupport = TRUE;
+    Info->SlotCount = 1;
```
petrutlucian94, line 313: I was a bit confused about the slot field but then I found those docs: https://docs.microsoft.com/en-us/windows-hardware/drivers/storage/upgrading-firmware-for-an-nvme-device

So we're having a virtual device, not using any firmware. Can we skip tests that do not apply in our case instead of cluttering the code with unused stuff?

driver/wnbd.inf
aserdean: Ooops.
Do you want me to add it to the commit message?

driver/scsi_driver_extensions.c
aserdean: I agree this is a bit mind numbing, but his applies just to one particular test on the 2019 playlist.

Unfortunately that test can't be skipped.

We could apply for an errata:
https://docs.microsoft.com/en-us/windows-hardware/test/hlk/user/windows-hardware-lab-kit-filters
but I have no idea about the resolution and how long that would take.

I added the implementation so we can have a clean run.

We can revert the commit later if we apply and get an errata for it.

PR #7, opened 2020-06-24
Add read-only mode

petrutlucian94:
This change will allow passing NBD flags. For now, we're only using
the read-only flag. In the future, we'll also detect and use
TRIM support.

### DIFF
--- a/driver/rbd_protocol.h
+++ b/driver/rbd_protocol.h
@@ -7,13 +7,24 @@
 #ifndef RBD_PROTOCOL_H
 #define RBD_PROTOCOL_H 1
 
-#include "common.h"
-
 #define NBD_REQUEST_MAGIC 0x25609513
 #define NBD_REPLY_MAGIC   0x67446698
 
 #define NBD_REQUEST_MAGIC 0x25609513
 #define NBD_REPLY_MAGIC   0x67446698
+
+/* values for flags field, these are server interaction specific. */
+#define NBD_FLAG_HAS_FLAGS  (1 << 0) /* nbd-server supports flags */
+#define NBD_FLAG_READ_ONLY  (1 << 1) /* device is read-only */
+#define NBD_FLAG_SEND_FLUSH (1 << 2) /* can flush writeback cache */
+#define NBD_FLAG_SEND_FUA   (1 << 3) /* send FUA (forced unit access) */
+/* there is a gap here to match userspace */
+#define NBD_FLAG_SEND_TRIM  (1 << 5) /* send trim/discard */
+#define NBD_FLAG_CAN_MULTI_CONN (1 << 8) /* Server supports multiple connections per export. */
+
+/* values for cmd flags in the upper 16 bits of request type */
+#define NBD_CMD_FLAG_FUA    (1 << 16) /* FUA (forced unit access) op */
+
 
 enum {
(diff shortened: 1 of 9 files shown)


driver/rbd_protocol.h
```diff
@@ -7,14 +7,25 @@
...
 #define NBD_REPLY_MAGIC   0x67446698
 
+/* values for flags field, these are server interaction specific. */
+#define NBD_FLAG_HAS_FLAGS  (1 << 0) /* nbd-server supports flags */
+#define NBD_FLAG_READ_ONLY  (1 << 1) /* device is read-only */
+#define NBD_FLAG_SEND_FLUSH (1 << 2) /* can flush writeback cache */
+#define NBD_FLAG_SEND_FUA   (1 << 3) /* send FUA (forced unit access) */
+/* there is a gap here to match userspace */
+#define NBD_FLAG_SEND_TRIM  (1 << 5) /* send trim/discard */
+#define NBD_FLAG_CAN_MULTI_CONN (1 << 8) /* Server supports multiple connections per export. */
```
aserdean, line 23: I'm not sure if we should add server flags here.

driver/scsi_operation.c
```diff
@@ -229,6 +229,7 @@ UCHAR
 WnbdSetModeSense(_In_ PVOID Data,
                  _In_ PCDB Cdb,
                  _In_ ULONG MaxLength,
+                 _In_ BOOLEAN ReadOnly,
```
aserdean, line 232: It might be better to pass `PSCSI_DEVICE_INFORMATION` here

driver/util.c
```diff
@@ -315,6 +315,26 @@ WnbdProcessDeviceThreadRequests(_In_ PSCSI_DEVICE_INFORMATION DeviceInformation)
...
+                case SCSIOP_SYNCHRONIZE_CACHE16:
+                    Element->Srb->DataTransferLength = 0;
+                    Element->Srb->SrbStatus = SRB_STATUS_INVALID_REQUEST;
+                    WNBD_LOG_LOUD("Write or flush requested on a read-only disk.",
+                                  Status, Element->Srb, Element->Tag);
+                    StorPortNotification(RequestComplete,
+                                         Element->DeviceExtension,
+                                         Element->Srb);
+                    ExFreePool(Element);
+                    return;
```
aserdean, line 334: I think you want `continue` instead of `return` here.

aserdean:
Testing looks good

driver/rbd_protocol.h
petrutlucian94: We should. That's the way in which the server advertises supported features. For example, if NBD_FLAG_SEND_TRIM isn't set, we shouldn't send trim requests.

driver/scsi_operation.c
petrutlucian94: Well, we only need one flag. It should be fine for now.

driver/util.c
petrutlucian94: You're right, thanks a lot for bringing this up!

driver/rbd_protocol.h
aserdean: Thanks for the explanation.

aserdean:
Thank you for addressing the comments.

PR #8, opened 2020-06-26
Fix assignment and checks for NBD flags

aserdean:
Signed-off-by: Alin Gabriel Serdean <aserdean@cloudbasesolutions.com>


petrutlucian94:
Works as expected, thanks!

PR #9, opened 2020-07-02
nit: Update the readme to use the standard NBD port

petrutlucian94:
Our readme sample is configuring the NBD server to use port 9000.
We'll switch to the IANA assigned NBD port, 10809.


aserdean:
Thanks for updating the documentation.

PR #12, opened 2020-08-28
Update WNBD API

petrutlucian94:
Update WNBD API

We're updating the WNBD driver interface, adding an alternative to the NBD protocol, which doesn't perform well enough (mostly due to WSK, especially on Windows Server 2016).

The new interface is also meant to be more consistent, easier to extend and less NBD centric.

To make it easier to consume, we're adding a library (wnbd.dll). It provides a low level API (the *Ioctl* functions), along with a high level API that can handle request dispatching.


aserdean:
Thank you for all the changes. The performance increase is substantial and overall stability looks good.

I managed to do a first round of tests on 2012, 2012 R2, 2016 and 2019.

There might be a few more things to iron out but let's address them in future changes and get the bulk in.

PR #13, opened 2020-08-31
Fix max block count

### LINKED ISSUE #11: Do not support device larger than 2TB
The block size default is 512.
When device larger than 2TB, `BlockCount` will overflow.

https://github.com/cloudbase/wnbd/blob/75ddd4832908153808a60723d38e988a6bfa41f0/driver/scsi_operation.c#L59

CzBiX:
Fix #11.

### DIFF
--- a/driver/scsi_operation.c
+++ b/driver/scsi_operation.c
@@ -17,33 +17,43 @@ UCHAR
 WnbdReadCapacity(_In_ PSCSI_DEVICE_INFORMATION Info,
                  _In_ PSCSI_REQUEST_BLOCK Srb,
                  _In_ PCDB Cdb,
-                 _In_ UINT64 BlockSize,
+                 _In_ UINT BlockSize,
                  _In_ UINT64 BlockCount)
 {
     WNBD_LOG_LOUD(": Enter");
     ASSERT(Srb);
     ASSERT(Cdb);
-    WNBD_LOG_INFO("Using BlockCount: %llu", BlockCount);
+    WNBD_LOG_INFO("Using BlockSize: %u, BlockCount: %llu", BlockSize, BlockCount);
 
     PVOID DataBuffer = SrbGetDataBuffer(Srb);
     ULONG DataTransferLength = SrbGetDataTransferLength(Srb);
-    UCHAR SrbStatus = SRB_STATUS_DATA_OVERRUN;
+    UCHAR SrbStatus = SRB_STATUS_INVALID_REQUEST;
 
-    if (0 == DataBuffer) {
+    if (0 == DataBuffer || 0 == BlockCount) {
+        SrbStatus = SRB_STATUS_INTERNAL_ERROR;
         goto Exit;
     }
+    UINT64 maxLba = BlockCount - 1;
 
     RtlZeroMemory(DataBuffer, DataTransferLength);
 
-    switch (Cdb->AsByte[0])
+    switch (Cdb->CDB10.OperationCode)
     {
     case SCSIOP_READ_CAPACITY:
         {
         if (sizeof(READ_CAPACITY_DATA) > DataTransferLength) {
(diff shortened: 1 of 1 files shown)


driver/scsi_operation.c
```diff
@@ -30,55 +30,85 @@ UCHAR
...
         {
         if (sizeof(READ_CAPACITY_DATA) > DataTransferLength) {
             goto Exit;
         }
+
         PREAD_CAPACITY_DATA ReadCapacityData = DataBuffer;
-        REVERSE_BYTES_4(&ReadCapacityData->LogicalBlockAddress, &BlockCount);
+
+        if (maxLba >= MAXULONG) {
+            ReadCapacityData->LogicalBlockAddress = MAXULONG;
```
petrutlucian94, line 64: 0xffffffff will be the same regardless of endianness, so this looks good.

driver/scsi_operation.c
```diff
@@ -30,55 +30,85 @@ UCHAR
...
     ULONG DataTransferLength = SrbGetDataTransferLength(Srb);
-    UCHAR SrbStatus = SRB_STATUS_DATA_OVERRUN;
+    UCHAR SrbStatus = SRB_STATUS_INVALID_REQUEST;
 
-    if (0 == DataBuffer) {
+    if (0 == DataBuffer || 0 == BlockCount) {
+        SrbStatus = SRB_STATUS_INTERNAL_ERROR;
         goto Exit;
     }
+    UINT64 maxLba = BlockCount - 1;
```
petrutlucian94, line 49: Good catch.

driver/scsi_operation.c
```diff
@@ -30,55 +30,85 @@ UCHAR
...
+        ULONG ReturnDataLength = 0;
         PREAD_CAPACITY16_DATA ReadCapacityData16 = DataBuffer;
-        REVERSE_BYTES_8(&ReadCapacityData16->LogicalBlockAddress, &BlockCount);
-        REVERSE_BYTES_8(&ReadCapacityData16->BytesPerBlock, &BlockSize);
-        if (CHECK_NBD_SEND_TRIM(Info->UserEntry->NbdFlags)) {
-            ReadCapacityData16->LBPME = 1;
+        REVERSE_BYTES_8(&ReadCapacityData16->LogicalBlockAddress, &maxLba);
+        REVERSE_BYTES_4(&ReadCapacityData16->BytesPerBlock, &BlockSize);
+
+        if (DataTransferLength >= (ULONG)FIELD_OFFSET(READ_CAPACITY16_DATA, Reserved3)) {
```
petrutlucian94, line 90: Supporting READ_CAPACITY_DATA_EX seems like a good idea.

driver/scsi_operation.c
```diff
@@ -30,55 +30,85 @@ UCHAR
...
+            if (CHECK_NBD_SEND_TRIM(Info->UserEntry->NbdFlags)) {
```
petrutlucian94, line 91: This will have to be replaced with the following, as per recent changes:

`if (Info->UserEntry->Properties.Flags.UnmapSupported) {`

petrutlucian94 (changes requested):
Looks good, thanks a lot for sending this fix! Could you please rebase the PR so that we can merge it?

driver/scsi_operation.c
CzBiX: Yes, we have to support `READ_CAPACITY_DATA_EX`, the buffer size of system sizing request always not big enough to put `READ_CAPACITY16_DATA`.

petrutlucian94:
@CzBiX one last thing before you rebase, can you please add a "Signed-off-by:" line to the commit message? I'm providing a contributor guide as we speak, similar to the Ceph one. That line will imply the license agreement.

PR #18, opened 2020-09-15
Code analysis

aserdean:
This PR addresses code analysis warnings found in `driver` project.

Adds a new configuration called `Analysis` so the Microsoft code analyzer is ran, with the default rule set, when building the project.

Also update appveyor matrix so we can see potential defects.

A sample output can be viewed here:
https://ci.appveyor.com/project/aserdean/wnbd-1/build/job/f382y7rx6ygidm91#L251

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,6 +4,7 @@ branches:
   only:
   - master
 configuration:
+  - Analyze
   - Debug
   - Release
 clone_folder: C:\wnbd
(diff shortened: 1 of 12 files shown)


petrutlucian94:
Wouldn't it be better to just enable ``<RunCodeAnalysis>true</RunCodeAnalysis>`` for the existing configurations (Release/Debug)?

aserdean:
We could, but it greatly increases build times (3 times - https://ci.appveyor.com/project/aserdean/wnbd-1/builds/35215661), hence the new configuration.

One minute and a half isn't that bad though.

petrutlucian94:
Makes sense, let's keep it as a separate build configuration then.

PR #20, opened 2020-09-16
Add versioning

petrutlucian94:
We'll use semantic versioning, which will be set through git tags.

For now, a script will be run as a pre-build event, parsing the git
tag and generating version.h.

wnbd-client -v will return the wnbd-client, the libwnbd as well as
the driver version.

We're using the "git describe" short version, which includes the
commit id and offset for untagged commits. If needed, we can easily
add other version info (e.g. build type, architecture, etc).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,5 @@ tags
 /vstudio/x64/
 /vstudio/sdv/
 /vstudiox64/
+
+include/version.h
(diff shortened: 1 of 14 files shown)


aserdean (approved):
Signed-off-by: Alin Gabriel Serdean <aserdean@cloudbasesolutions.com>

PR #22, opened 2020-09-24
Refactor WNBD device handling

petrutlucian94:
Refactor WNBD device handling

At the moment, we have about 4 overlapping (providing the same fields),
highly interconnected structures that describe a device, as well as two device
lists. Locking isn't ideal either, we're using more locks than we need.
Operations such as creating, listing or removing devices are holding exclusive
locks more than they should.

At the same time, the device cleanup workflow is convoluted and doesn't
guarantee that all the pending requests are completed by the time we cleanup
disk devices or the entire extension, which often leads to memory leaks.

This refactoring is meant to greatly simplify device handling, making the driver
easier to maintain and also avoid a few memory leaks. Here are the changes that
we're going to make:

* drop unneeded (but heavily used) structures: WNBD_LU_EXTENSION,
  GLOBAL_INFORMATION
* merge structures that provide device info: USER_ENTRY and
  SCSI_DEVICE_INFORMATION are merged into the existing WNBD_SCSI_DEVICE struct
* use reference counting (via Rundown Protection objects) at the disk device
  as well as extension level. Before accessing a device, a reference must be
  acquired (handled by WnbdFindDevice* functions), ensuring that it won't be
  cleaned up while still being accessed. At the same time, the device reference
  count will ensure that all devices are removed before we clean up the
  extension.

  Rundown Protection objects are highly convenient, providing locking as well.
  We can wait for all pending operations to complete, and once we obtain the
  lock, no other operations will be allowed.
* we'll use one monitor/cleanup thread per device instead of the global one.
  which will allow cleaning up devices in parallel. Having all the device
  disconnect and cleanup logic in one place makes it easier to follow and less
  error prone. It also makes it easier to tell the IRQL used when disconnecting
  the device. We can also avoid device cleanup race conditions without having
  to use some of the existing (global) locks.
* at the moment, there are various way in which we attempt to disconnect or
  cleanup a device (e.g. setting WNBD_SCSI_DEVICE.Missing,
  WNBD_SCSI_DEVICE.ReportedMissing, PSCSI_DEVICE_INFORMATION.HardTermiateDevice
  or calling DisconnectConnection, trigger the device cleanup thread).
  that seems overly complicated. this change will allow removing a device
  by either signaling the device terminate event or the global device terminate
  event (which in turn will remove all devices).

  In order to synchronously remove a device, the caller can hold a device
  reference, obtain a device cleanup thread reference, signal the device
  monitor thread and then wait for it to complete.

  When cleaning up all the devices, all we have to do is signal the global
  device terminate event and then wait for the device reference count to reach
  0.
* simplify outstanding IO tracking and move the counter to WNBD_DRV_STATS
* simplify SRB status reporting, adding CompleteRequest
* simplify socket cleanup helpers and fix incorrect naming (
  currently DisconnectConnection closes the socket and CloseConnection
  tries to gracefully disconnect it)
* avoid using void pointers when not needed, it makes the interface difficult
  to follow
* move NBD dispatchers to nbd_dispatch.c


driver/scsi_driver_extensions.h
```diff
@@ -8,48 +8,78 @@
...
 
-    UNICODE_STRING					DeviceInterface;
-    LIST_ENTRY						DeviceList;
-    KSPIN_LOCK						DeviceListLock;
-    ERESOURCE                       DeviceResourceLock;
+    UNICODE_STRING                    DeviceInterface;
+    LIST_ENTRY                        DeviceList;
+    KSPIN_LOCK                        DeviceListLock;
+    LONG                              DeviceCount;
+    ERESOURCE                         DeviceResourceLock;
```
petrutlucian94, line 20: TODO: rename as DeviceCreationLock.

driver/util.c
```diff
@@ -15,458 +16,253 @@
...
 
-    ExReleaseResourceLite(&ScsiInfo->GlobalInformation->ConnectionMutex);
-    KeLeaveCriticalRegion();
+    // The rundown protection is a device reference count. We're going to wait
+    // for them to be removed after signaling the global device removal event.
+    ExWaitForRundownProtectionRelease(&DeviceExtension->RundownProtection);
 
-    ExFreePool(ScsiInfo);
-    ScsiInfo = NULL;
+    // TODO: check if we can destroy uninitialized WSK.
```
aserdean, line 93: As is, we can't. We can change `KsDestroy` though.

driver/scsi_driver_extensions.h
```diff
@@ -8,48 +8,78 @@
...
+    USHORT                      Target;
+    USHORT                      Lun;
+
+    PINQUIRYDATA                InquiryData;
+
+    INT                         NbdSocket;
+    INT                         SocketToClose;
+    ERESOURCE                   SocketLock;
+
+    // TODO: rename as PendingReqListHead
```
aserdean, line 45: Since we are changing the structures, it may be a good time to rename.

driver/scsi_driver_extensions.h
```diff
@@ -8,48 +8,78 @@
...
+    LIST_ENTRY                  RequestListHead;
+    KSPIN_LOCK                  RequestListLock;
+
+    // TODO: rename as SubmittedReqListHead
```
aserdean, line 49: Since we are changing the structures, it may be a good time to rename.

driver/wnbd_dispatch.c
```diff
@@ -83,7 +68,7 @@ NTSTATUS LockUsermodeBuffer(
 
 NTSTATUS WnbdDispatchRequest(
     PIRP Irp,
-    PSCSI_DEVICE_INFORMATION DeviceInfo,
+    PWNBD_SCSI_DEVICE Device,
     PWNBD_IOCTL_FETCH_REQ_COMMAND Command)
 {
     // TODO: reject request when using NBD.
```
aserdean, line 74: Can these be dropped now?

driver/nbd_dispatch.c
```diff
@@ -0,0 +1,323 @@
...
+        }
+    }
+    if (Reply.Error) {
+        // TODO: do we care about the actual error?
+        WNBD_LOG_INFO("NBD reply contains error: %llu", Reply.Error);
+        Element->Srb->DataTransferLength = 0;
+        Element->Srb->SrbStatus = SRB_STATUS_ABORTED;
+    }
+    else {
+        // TODO: rename ReadLength to DataLength
```
aserdean, line 303: Maybe a good time to do the rename?

aserdean (commented):
Thanks for setting this up.
I found some nits, looking over the changes.

I will test the changes and let you know if I find something.

driver/wnbd_dispatch.c
petrutlucian94: Definitely, I forgot to remove the TODO. Thanks!

petrutlucian94 (commented):
Thanks for feedback!

aserdean (changes requested):
Just a few more nits. Will add testing results soon.

driver/nbd_dispatch.c
```diff
@@ -0,0 +1,323 @@
...
+        }
+    }
+    if (Reply.Error) {
+        // TODO: do we care about the actual error?
+        WNBD_LOG_INFO("NBD reply contains error: %llu", Reply.Error);
+        Element->Srb->DataTransferLength = 0;
+        Element->Srb->SrbStatus = SRB_STATUS_ABORTED;
+    }
+    else {
+        // TODO: rename DataLength to DataLength
```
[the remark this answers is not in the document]
petrutlucian94: Whops, thanks.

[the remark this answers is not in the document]
petrutlucian94: Whops, thanks.

driver/scsi_driver_extensions.h
```diff
@@ -8,48 +8,81 @@
...
+    LIST_ENTRY                  SubmittedReqListHead;
+    KSPIN_LOCK                  SubmittedReqListLock;
+
+    KSEMAPHORE                  DeviceEvent;
+    PVOID                       DeviceRequestThread;
+    PVOID                       DeviceReplyThread;
+    PVOID                       DeviceMonitorThread;
+    BOOLEAN                     HardTerminateDevice;
+    BOOLEAN                     SoftTerminateDevice;
+    KEVENT                      TerminateEvent;
```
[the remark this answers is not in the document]
petrutlucian94: Makes sense.

driver/nbd_dispatch.c
```diff
@@ -0,0 +1,323 @@
...
+    WNBD_LOG_LOUD(": Enter");
+    ASSERT(Device);
+    ASSERT(Element);
+    ULONG StorResult;
+    PVOID Buffer;
+    NTSTATUS Status = STATUS_SUCCESS;
+
+    StorResult = StorPortGetSystemAddress(Element->DeviceExtension, Element->Srb, &Buffer);
+    if (STOR_STATUS_SUCCESS != StorResult) {
+        Status = SRB_STATUS_INTERNAL_ERROR;
```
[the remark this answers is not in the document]
petrutlucian94: Sure.

driver/wnbd_dispatch.c
```diff
@@ -164,65 +147,53 @@ NTSTATUS WnbdDispatchRequest(
...
+                InterlockedDecrement64(&Device->Stats.UnsubmittedIORequests);
                 Status = STATUS_BUFFER_TOO_SMALL;
                 goto Exit;
             }
 
             PVOID SrbBuffer;
             if (StorPortGetSystemAddress(Element->DeviceExtension,
                                          Element->Srb, &SrbBuffer)) {
-                // TODO: consider moving this part to a helper function.
                 Element->Srb->SrbStatus = SRB_STATUS_INTERNAL_ERROR;
```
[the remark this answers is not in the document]
petrutlucian94: Yep. The only issue that I see is the log level, we might want to avoid WNBD_LOG_ERROR in the hot path not only for performance reasons, but to avoid spamming the logs (especially if we're to use ETW).

driver/scsi_driver_extensions.h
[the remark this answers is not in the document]
petrutlucian94: I did that but in the next PR to avoid a bunch of merge conflicts.

aserdean:
The fixes found while are testing are in PR: https://github.com/cloudbase/wnbd/pull/23.

I'm closing this PR and merging it.

PR #23, opened 2020-10-01
Add soft device removal using PnP

petrutlucian94:
Add soft device removal using PnP

We'll update libwnbd to allow notifying the PnP stack before asking
the driver to disconnect a device. PnP subscribers (e.g. filesystem
or volume managers) can veto device removal requests until all data
has been flushed.

We'll add a configurable timeout and a hard removal fallback option.
Since we're still in beta phase, we're going to go ahead and change
the device remove API.

On the driver side, we're updating the PnP capabilities. Disabling
"SurpriseRemovalOk" says that the device is stateful (we're holding
userspace connections) and the driver must be notified when a device
is removed. At the same time, we're enabling EjectSupported.

We're adding a "show" command, making it easier for libwnbd to
retrieve the disk serial number based on the instance name, which
is then needed when iterating over PnP device lists.

Ideally, our driver should react to PnP device removal events.
TL;DR: at the moment it doesn't, so we'll need an additional
IOCTL_WNBD_UNMAP request. The long story is that we can receive
PnP remove events either through IRP_MN_REMOVE_DEVICE or StorRemoveDevice.
The latter would be preferred because it should provide us the device
extension reference as well as the SCSI disk address. Unfortunately,
we're not receiving those, just getting IRP_MN_REMOVE_DEVICE,
which only provides the PDO, which seems to have the wrong extension
address.

To make it easier to handle PnP removal events, we're moving the
synchronous device removal logic to WnbdDisconnectSync, which
will be the counterpart of the existing WnbdDisconnectAsync.

Depends-On: https://github.com/cloudbase/wnbd/pull/22


aserdean:
Thanks a lot for all the work, testing looks good!

Thank you for addressing the comments and including the fixes!

PR #24, opened 2020-10-02
userspace: update logging

petrutlucian94:
userspace: update logging

We allow the library consumer to provide a logger function.
This is really convenient because the consumer can choose how those
messages will be handled (e.g. can use ETW or simple file logging,
can format the messages, etc).

The main issue is that the logger is tied to the WNBD_DEVICE
structure, which limits its usefulness since we can't use it
in functions that don't accept this parameter.

This change will add a static logger function pointer, allowing
the consumer to override it using a setter.

This commit also adds log messages to all the functions that couldn't
use logging before.

Depends-On: https://github.com/cloudbase/wnbd/pull/23

### DIFF
--- a/include/wnbd.h
+++ b/include/wnbd.h
@@ -1,6 +1,7 @@
 #ifndef WNBD_SHARED_H
 #define WNBD_SHARED_H
 
+#include <windows.h>
 #include <cfgmgr32.h>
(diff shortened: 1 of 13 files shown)


aserdean:

PR #26, opened 2020-10-06
driver: unmap/flush/FUA support

petrutlucian94:
driver: unmap/flush/FUA support

Those features are already supported by libwnbd and our nbd
client, we'll need to update our IO dispatcher to propagate
them.

### DIFF
--- a/driver/wnbd_dispatch.c
+++ b/driver/wnbd_dispatch.c
@@ -126,7 +126,6 @@ NTSTATUS WnbdDispatchRequest(
             continue;
         }
 
-        // TODO: consider moving this part to a helper function.
         PSRB_QUEUE_ELEMENT Element = CONTAINING_RECORD(RequestEntry, SRB_QUEUE_ELEMENT, Link);
         Element->Tag = InterlockedIncrement64(&(LONG64)RequestHandle);
         Element->Srb->DataTransferLength = 0;
@@ -136,15 +135,8 @@ NTSTATUS WnbdDispatchRequest(
         WnbdRequestType RequestType = ScsiOpToWnbdReqType(Cdb->AsByte[0]);
         WNBD_LOG_LOUD("Processing request. Address: %p Tag: 0x%llx Type: %d",
                       Element->Srb, Element->Tag, RequestType);
-        // TODO: check if the device supports the requested operation
-        switch(RequestType) {
-        case WnbdReqTypeRead:
-        case WnbdReqTypeWrite:
-            Request->RequestType = RequestType;
-            Request->RequestHandle = Element->Tag;
-            break;
-        // TODO: flush/unmap
-        default:
+
+        if (!ValidateScsiRequest(Device, Element)) {
             Element->Srb->DataTransferLength = 0;
             Element->Srb->SrbStatus = SRB_STATUS_INVALID_REQUEST;
             CompleteRequest(Device, Element, TRUE);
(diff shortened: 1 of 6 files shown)


libwnbd/wnbd_ioctl.cpp
```diff
@@ -95,10 +99,17 @@ DWORD WnbdOpenAdapterEx(PHANDLE Handle, PDEVINST CMDeviceInstance)
...
-            "driver is installed. Error: %d. Error message: %s",
-            ErrorCode, win32_strerror(ErrorCode).c_str());
+        if (ErrorCode == ERROR_ACCESS_DENIED) {
+            LogError(
+                "Could not open WNBD adapter device. Access denied, try "
+                "using an elevated command prompt.");
+        }
+        else {
+            LogError(
+                "Could not oped WNBD adapter device. Please make sure that "
```
aserdean, line 109: nit: open?

driver/wnbd_dispatch.c
```diff
@@ -190,6 +189,35 @@ NTSTATUS WnbdDispatchRequest(
 
             RtlCopyMemory(Buffer, SrbBuffer, Element->DataLength);
             break;
+        case WnbdReqTypeFlush:
+            Request->Cmd.Flush.BlockAddress =
+                Element->StartingLbn / DevProps->BlockSize;;
```
aserdean, line 194: nit: s/;;/;/

libwnbd/wnbd_ioctl.cpp
```diff
@@ -95,10 +99,17 @@ DWORD WnbdOpenAdapterEx(PHANDLE Handle, PDEVINST CMDeviceInstance)
         *CMDeviceInstance = DevInfoData.DevInst;
     }
     else {
-        LogError(
-            "Could not oped WNBD device. Please make sure that the "
-            "driver is installed. Error: %d. Error message: %s",
-            ErrorCode, win32_strerror(ErrorCode).c_str());
+        if (ErrorCode == ERROR_ACCESS_DENIED) {
+            LogError(
+                "Could not open WNBD adapter device. Access denied, try "
```
aserdean, line 104: Should the message be shortened to `Access denied, using an elevated command prompt.`?

libwnbd/wnbd_ioctl.cpp
```diff
@@ -95,10 +99,17 @@ DWORD WnbdOpenAdapterEx(PHANDLE Handle, PDEVINST CMDeviceInstance)
         *CMDeviceInstance = DevInfoData.DevInst;
     }
     else {
-        LogError(
-            "Could not oped WNBD device. Please make sure that the "
-            "driver is installed. Error: %d. Error message: %s",
-            ErrorCode, win32_strerror(ErrorCode).c_str());
+        if (ErrorCode == ERROR_ACCESS_DENIED) {
```
aserdean, line 102: `win32_strerror` translates the error code using `FormatMessage` which should provide a meaningful error though.

I understand the need to treat this error differently but I'm not really in favor of it.

aserdean (changes requested):

libwnbd/wnbd_ioctl.cpp
petrutlucian94: Whops, thanks.

libwnbd/wnbd_ioctl.cpp
petrutlucian94: It's not really the same thing, "Access denied, using an elevated command prompt." suggests that we're automatically using an elevated prompt as a fallback, which is confusing.

driver/wnbd_dispatch.c
petrutlucian94: Whops, thanks.

libwnbd/wnbd_ioctl.cpp
petrutlucian94: FormatMessage will just return "Access denied" without any other details. Using unelevated prompts is quite common and most people wouldn't figure out right away the cause of the "Access denied" error, so I really think we should add a more explicit message. There's no downside to it and we'll spare people from some wasted time debugging this, maybe even avoid some bug reports.

driver/wnbd_dispatch.c
aserdean: nit: s/;;/;/g

The first occurrence from `WnbdReqTypeRead` got fixed but there is still one in `WnbdReqTypeFlush`.

PR #29, opened 2020-10-15
Add new logging facilities and the ability to install and uninstall the driver via client

aserdean:
This PR adds two new logging facilities: one for WPP and one for ETW.

WPP tracing is enabled only on Debug builds for the moment and different tools
can be used used to capture its output.

ETW tracing is enabled on all builds.
To start a ETW trace session one can use:
`tracelog -start WNBDEventdrv -guid #FFACC4E7-C115-4FE2-9D3C-80FAE73BAB91 -f WNBDEventdrv.etl`

To stop:
`tracelog -stop WNBDEventdrv`

To display the trace use:
`tracerpt WNBDEventdrv.etl`

The `reinstall.ps1` script has been updated to install and remove our defined events.

Two new commands are added to `wnbd-client`: `install-driver <InfFileLocation>` and `uninstall-driver`.

The predefined hardware ID used for installation and lookup is `root\wnbd`.

When using the uninstall command please note that the OEM information is also marked for removal.

### DIFF
--- a/include/wnbd.h
+++ b/include/wnbd.h
@@ -9,6 +9,7 @@
 
 #include <windows.h>
 #include <cfgmgr32.h>
+#include <SetupAPI.h>
 
 #include "wnbd_ioctl.h"
 
@@ -21,6 +22,9 @@ extern "C" {
 #define WNBD_LOG_MESSAGE_MAX_SIZE 4096
 #define WNBD_DEFAULT_RM_TIMEOUT_MS 30 * 1000
 #define WNBD_DEFAULT_RM_RETRY_INTERVAL_MS 2000
+
+#define WNBD_HARDWAREID "root\\wnbd"
+#define WNBD_HARDWAREID_LEN sizeof(WNBD_HARDWAREID)
 
 typedef enum
 {
@@ -272,6 +276,8 @@ DWORD WnbdOpenAdapter(PHANDLE Handle);
 DWORD WnbdOpenAdapterEx(PHANDLE Handle, PDEVINST CMDeviceInstance);
 DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance);
 DWORD WnbdIoctlPing(HANDLE Adapter, LPOVERLAPPED Overlapped);
+DWORD WnbdUninstallDriver(PBOOL RebootRequired, BOOLEAN Fprce);
+DWORD WnbdInstallDriver(CONST CHAR* FileName, PBOOL RebootRequired);
 
 // The "Overlapped" parameter used by WnbdIoctl* functions allows
 // asynchronous calls. If NULL, a valid overlapped structure is
--- a/libwnbd/wnbd_ioctl.cpp
+++ b/libwnbd/wnbd_ioctl.cpp
@@ -8,15 +8,20 @@
 #include <windef.h>
 #include <winioctl.h>
 #include <winioctl.h>
+#include <newdev.h>
 #include <ntddscsi.h>
+#include "Shlwapi.h"
 #include <setupapi.h>
 #include <string.h>
+#include <infstr.h>
 
 #include "wnbd.h"
 #include "wnbd_log.h"
 #include "utils.h"
 
 #pragma comment(lib, "Setupapi.lib")
+#pragma comment(lib, "Newdev.lib")
+#pragma comment(lib, "Shlwapi.lib")
 
 #define STRING_OVERFLOWS(Str, MaxLen) (strlen(Str + 1) > MaxLen)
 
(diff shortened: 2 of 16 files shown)


vstudio/reinstall.ps1
```diff
@@ -31,8 +31,10 @@ if ($status -eq $false) {
 }
 
 & $devconBin remove "root\wnbd"
+wevtutil um wnbdevents.xml
```
petrutlucian94, line 34: Let's add the "$scriptLocation\" prefix. This script may be called from somewhere else. We might define a path such as "$wnbdEvents" and add it to line 8.

vstudio/wnbdevents.xml
```diff
@@ -0,0 +1,58 @@
...
+    xmlns:win="http://manifests.microsoft.com/win/2004/08/windows/events"
+    xmlns:xs="http://www.w3.org/2001/XMLSchema"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://schemas.microsoft.com/win/2004/08/events eventman.xsd">
+    <instrumentation>
+        <events>
+            <provider
+                guid="{FFACC4E7-C115-4FE2-9D3C-80FAE73BAB91}"
+                messageFileName="%WINDIR%\System32\drivers\wnbd.sys"
+                name="WinBD"
```
petrutlucian94, line 13: Let's stick to wnbd to match the project name.

vstudio/wnbdevents.xml
```diff
@@ -0,0 +1,58 @@
...
+                        value="1"/>
+                </events>
+            </provider>
+        </events>
+    </instrumentation>
+    <localization xmlns="http://schemas.microsoft.com/win/2004/08/events">
+        <resources culture="en-US">
+            <stringTable>
+                <string id="Event.EventMessage"
+                        value="WinBD Loaded"/>
```
petrutlucian94, line 54: Can we use the actual message instead? Otherwise all log entries will contain the "WinBD Loaded" message, which is confusing.

vstudio/wnbdevents.xml
aserdean: Sure.

vstudio/reinstall.ps1
aserdean: Whoops.

include/wnbd.h
```diff
@@ -16,6 +17,9 @@ extern "C" {
 #define WNBD_DEFAULT_RM_TIMEOUT_MS 30 * 1000
 #define WNBD_DEFAULT_RM_RETRY_INTERVAL_MS 2000
 
+#define WNBD_HARDWAREID "root\\wnbd"
+#define WNBD_HARDWAREID_LEN 10
```
petrutlucian94, line 21: nit: I guess we can just use sizeof(WNBD_HARDWAREID)

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,245 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
     return Status;
 }
 
+DWORD GetDrvDetailData(HDEVINFO DevHandle, PSP_DEVINFO_DATA DevInfoData,
```
petrutlucian94, line 140: nit: let's rename DevHandle to AdapterHandle or AdapterDevHandle, underlining the fact that this isn't actually a disk device.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,245 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    DWORD GLA = ERROR_SUCCESS;
+
+    InstallParams.cbSize = sizeof InstallParams;
+    if (!SetupDiGetDeviceInstallParamsA(DevHandle, DevInfoData, &InstallParams)) {
+        GLA = GetLastError();
+        goto exit;
+    }
+
+    InstallParams.FlagsEx |= DI_FLAGSEX_INSTALLEDDRIVER | DI_FLAGSEX_ALLOWEXCLUDEDDRVS;
+    if (!SetupDiSetDeviceInstallParamsA(DevHandle, DevInfoData, &InstallParams)) {
```
petrutlucian94, line 156: Why are we calling ``SetupDiSetDeviceInstallParamsA`` twice?

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,245 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+{
+    SP_DEVINSTALL_PARAMS_A InstallParams;
+    BOOL Cleanup = FALSE, Found = FALSE;
+    SP_DRVINFO_DATA_A DrvInfoData;
+    DrvInfoData.cbSize = sizeof DrvInfoData;
+    DWORD GLA = ERROR_SUCCESS;
+
+    InstallParams.cbSize = sizeof InstallParams;
+    if (!SetupDiGetDeviceInstallParamsA(DevHandle, DevInfoData, &InstallParams)) {
+        GLA = GetLastError();
```
petrutlucian94, line 151: nit: let's rename GLA to Status so that we're consistent with the rest of the code. GLA also sounds a bit cryptic.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,245 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    }
+
+    if (!SetupDiBuildDriverInfoList(DevHandle, DevInfoData, SPDIT_CLASSDRIVER)) {
+        GLA = GetLastError();
+        goto exit;
+    }
+    Cleanup = TRUE;
+
+    for (DWORD index = 0;
+        SetupDiEnumDriverInfoA(DevHandle, DevInfoData, SPDIT_CLASSDRIVER, index, &DrvInfoData);
```
petrutlucian94, line 168: I think we should move ``SetupDiEnumDriverInfoA`` inside the loop so that we can check the error code. If it's something other than ERROR_NO_MORE_ITEMS, we should log and propagate the error.

Same for ``SetupDiGetDriverInfoDetailA`` errors, I think we may want to log/propagate them.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,245 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    DWORD GLA = ERROR_SUCCESS;
+
+    GLA = GetDrvDetailData(DevHandle, DevInfoData, &DrvDetailData);
+    if (ERROR_SUCCESS != GLA) {
+        return GLA;
+    }
+
+    std::string SearchString(DrvDetailData.InfFileName);
+
+    if (!DiUninstallDevice(0, DevHandle, DevInfoData, 0, RebootRequired)) {
```
petrutlucian94, line 213: We should check the error code, maybe it failed for some other reason.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,245 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+            Found = TRUE;
+        }
+    }
+    if (!Found) {
+        goto exit;
+    }
+
+    GLA = ERROR_SUCCESS;
+
+exit:
```
petrutlucian94, line 256: I think we should avoid unnecessary ``goto``s. If the exit label does nothing other than returning a value, we should just use ``return`` instead of ``goto``.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,245 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    BOOL Found = FALSE;
+    DWORD GLA = ERROR_FILE_NOT_FOUND;
+
+    *DevHandle = SetupDiGetClassDevsA(0, 0, 0, DIGCF_ALLCLASSES | DIGCF_PRESENT);
+    if (INVALID_HANDLE_VALUE == *DevHandle) {
+        GLA = GetLastError();
+        goto exit;
+    }
+
+    for (DWORD I = 0; !Found && SetupDiEnumDeviceInfo(*DevHandle, I, DevInfoData); I++)
```
petrutlucian94, line 238: Should we check the error codes for ``SetupDiEnumDeviceInfo`` and ``SetupDiGetDeviceRegistryPropertyA``?

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,245 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    {
+        if (!SetupDiGetDeviceRegistryPropertyA(*DevHandle, DevInfoData, SPDRP_HARDWAREID, 0,
+            (PBYTE)TempBuf, 2046, 0)) {
```
petrutlucian94, line 241: Should we use ``sizeof(TempBuf) - 1`` instead of 2046?

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,245 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+            continue;
+        }
+
+        std::string TempString(TempBuf);
```
petrutlucian94, line 245: Should we just use strcasestr? Also, I assume we don't care about devices that have a path other than "root/wnbd".

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,245 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    HDEVINFO DevHandle = INVALID_HANDLE_VALUE;
+    SP_DEVINFO_DATA DevInfoData;
+    DevInfoData.cbSize = sizeof DevInfoData;
+    GLA = FindDriver(&DevHandle, &DevInfoData);
+    if (ERROR_SUCCESS != GLA) {
+        LogError("Driver not installed");
+        return GLA;
+    }
+    // TODO Lock mappings
+    // TODO Unmap all mappings
```
petrutlucian94, line 272: We can use: ``WnbdSetOption("NewMappingsAllowed", {.Type = WnbdOptBool, .Data.AsBool = FALSE }, FALSE)``

It should be really easy to iterate over the mappings using ``WnbdList`` and remove them using ``WnbdRemove``. I'd go with a hard remove.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,245 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    if (SearchString.length()) {
+        SetupUninstallOEMInfA(
+            SearchString.substr(SearchString.find_last_of("\\") + 1).c_str(), SUOI_FORCEDELETE, 0);
+    }
+
+    return GLA;
+}
+
+
+static DWORD FindDriver(HDEVINFO* DevHandle, SP_DEVINFO_DATA* DevInfoData)
```
petrutlucian94, line 226: The name is a bit confusing, we aren't searching for a driver but for an actual device.

petrutlucian94 (changes requested):

vstudio/wnbdevents.xml
aserdean: Whops forgot about the message will change it based on the event data.

We include the log level in the event data.

We could add an event based on the driver log level and check inside `WnbdLog` which event to send.
If we do the above another question would be which events would be logged when tracing is enabled and which are enabled by default.

petrutlucian94: The log level is already configurable at the WNBD level. I'd still prefer to properly set the log level on the entry, not just the payload. An error shouldn't show up in the event log as an info message.

I think we can set Event.Level just like we set Event.EventMessage: https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/adding-event-tracing-to-kernel-mode-drivers.

libwnbd/wnbd_ioctl.cpp
petrutlucian94: Ohh, right, sorry for this one.

libwnbd/wnbd_ioctl.cpp
petrutlucian94: Yep, but we may fail to locate it due to some unrelated error, even if the driver exists. In that case, the error should be logged and propagated.

aserdean: > 
> 
> Yep, but we may fail to locate it due to some unrelated error, even if the driver exists. In that case, the error should be logged and propagated.

`the error will be propagated via line 181 and logged at line 188 , because Found will be false.`

vstudio/wnbdevents.xml
aserdean: Unfortunately it can't, only the message is force to be overwritten. This is why I suggested we create an entry per type of log level.

Indeed the level of tracing is configurable at the driver level, but I was referring to the fact that no actual events show up in the event log at the moment, only on traces we start/stop.
To add events directly to event log we need to define them at different `opcodes`.

I was thinking to log only errors directly to the event log and the via traces. Do you agree?

aserdean: As discussed offline we are logging `WNBD_LVL_ERROR` and `WNBD_LVL_WARN` directly in event viewer and the rest will appear only when a trace is enabled on our ETW provider.

libwnbd/wnbd_ioctl.cpp
aserdean: Thanks for the pointers. I added the function: `HardRemoveAllDevices` for both of the `TODO` items.

vstudio/wnbdevents.xml
```diff
@@ -0,0 +1,72 @@
...
+                              inType="win:AnsiString"
+                              outType="xs:string"/>
+                    </template>
+                </templates>
+                <events>
+                    <event
+                        channel="SYSTEM"
+                        level="win:Error"
+                        message="$(string.ErrorEvent.EventMessage)"
+                        opcode="win:Info"
```
petrutlucian94, line 39: Please correct me if I'm wrong, but from what I can tell ``opcode`` is only used Windows versions older than Windows 7. Can we skip it? It's a bit confusing that we have ``opcode="win:Info"`` and ``level="win:Error".

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    }
+
+    if (!SetupDiBuildDriverInfoList(AdapterDevHandle, DevInfoData, SPDIT_CLASSDRIVER)) {
+        Status = GetLastError();
+        goto exit;
+    }
+    Cleanup = TRUE;
+
+    for (DWORD index = 0;
+        SetupDiEnumDriverInfoA(AdapterDevHandle, DevInfoData, SPDIT_CLASSDRIVER, index, &DrvInfoData);
```
petrutlucian94, line 168: This looked a bit odd to me at first, I didn't know why we're looping over a list of members and how to tell which is the right one.

Apparently SetupDiEnumDriverInfo lists all the drivers associated with a device. Are we sure there will be a single one, WNBD? If so, let's add an inline comment, something like: `` // SetupDiEnumDriverInfoA enumerates all the drivers associated with a device. We expect WNBD to be the only driver associated with the device``.

I'm wondering if we should throw a warning if by any chance there is more than one associated driver (maybe some improper cleanup?).

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    DWORD Status = 0;
+    PWNBD_CONNECTION_LIST TempList = NULL;
+
+    WNBD_OPTION_VALUE OptValue = { WnbdOptBool };
+    /* Disallow new mappings so we can remove all current mappings */
+    OptValue.Data.AsBool = FALSE;
+    Status = WnbdSetDrvOpt("NewMappingsAllowed", &OptValue, FALSE);
+    if (Status) {
+        LogError(
+            "WnbdSetDrvOpt Error: %d. Error message: %s", Status, win32_strerror(Status).c_str());
```
petrutlucian94, line 281: I think we can drop this error, WnbdIoctlSetDrvOpt will log it as well so we'd end up with the same error being logged twice.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+
+    Status = WnbdList(TempList, &BufferSize);
+    if (!BufferSize) {
+        goto exit;
+    }
+    TempList = (PWNBD_CONNECTION_LIST)calloc(1, BufferSize);
+    if (!TempList) {
+        Status = ERROR_NOT_ENOUGH_MEMORY;
+        LogError(
+            "calloc Error: %d. Error message: %s", Status, win32_strerror(Status).c_str());
```
petrutlucian94, line 293: nit: let's go with a more user friendly message. Something like: ``"Failed to allocate %d bytes.", BufferSize``

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+        goto exit;
+    }
+    Status = WnbdList(TempList, &BufferSize);
+    if (Status) {
+        LogError(
+            "WnbdList Error: %d. Error message: %s", Status, win32_strerror(Status).c_str());
```
petrutlucian94, line 299: ditto, I think we can drop the error.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    }
+    CHAR* InstanceName;
+    for (ULONG index = 0; index < TempList->Count; index++) {
+        InstanceName = TempList->Connections[index].Properties.InstanceName;
+        WNBD_REMOVE_OPTIONS RemoveOptions = { 0 };
+        RemoveOptions.Flags.HardRemove = TRUE;
+        Status = WnbdRemoveEx(InstanceName, &RemoveOptions);
+        if (Status) {
+            LogError(
+                "Could not remove device: %s. "
```
petrutlucian94, line 310: ditto about the error message.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+
+        if (strstr(TempBuf, WNBD_HARDWAREID) != NULL) {
+            Found = TRUE;
+            break;
+        }
+    }
+
+    if (!Found) {
+        Status = GetLastError();
+        LogInfo("Error: %d. Error message: %s", Status, win32_strerror(Status).c_str());
```
petrutlucian94, line 262: This message doesn't provide any context information so it would be very confusing. Let's add "Failed to locate device. Error: ...".

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    }
+
+    return ERROR_SUCCESS;
+}
+
+DWORD HardRemoveAllDevices()
+{
+    DWORD BufferSize = 0;
+    DWORD Status = 0;
+    PWNBD_CONNECTION_LIST TempList = NULL;
```
petrutlucian94, line 273: really minor nit: s/TempList/ConnList.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    if (ERROR_SUCCESS != Status) {
+        LogError(
+            "Could not remove driver. "
+            "Error: %d. Error message: %s", Status, win32_strerror(Status).c_str());
+    }
+    return Status;
+}
+
+
+static DWORD FindDevice(HDEVINFO* AdapterDevHandle, SP_DEVINFO_DATA* DevInfoData)
```
petrutlucian94, line 231: Nit: I think we should rename the function to FindWnbdAdapterDevice since we are explicitly searching for a WNBD adapter device.

Also, let's add an inline comment stating that ``// We're using a single WNBD adapter per host``.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+{
+    CHAR TempBuf[2048];
+    memset(TempBuf, 0, sizeof(TempBuf));
+    BOOL Found = FALSE;
+    DWORD Status = ERROR_FILE_NOT_FOUND;
+
+    *AdapterDevHandle = SetupDiGetClassDevsA(0, 0, 0, DIGCF_ALLCLASSES | DIGCF_PRESENT);
```
petrutlucian94, line 238: Should we use ``WNBD_GUID`` as ``WnbdOpenAdapterEx`` does? We could print a warning when there are multiple WNBD devices.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    return Status;
+}
+
+DWORD WnbdUninstallDriver(PBOOL RebootRequired)
+{
+    DWORD Status;
+    HDEVINFO AdapterDevHandle = INVALID_HANDLE_VALUE;
+    SP_DEVINFO_DATA DevInfoData;
+    DevInfoData.cbSize = sizeof DevInfoData;
+    Status = FindDevice(&AdapterDevHandle, &DevInfoData);
```
petrutlucian94, line 330: Just because there isn't a WNBD device doesn't mean that the driver is not installed. We should go ahead and actually remove the driver.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    if (ERROR_SUCCESS == FindDevice(&AdapterDevHandle, &DevInfoData)) {
+        LogError("Driver already installed");
+        Status = ERROR_DUPLICATE_FOUND;
+        goto exit;
+    }
+
+    Status = InstallDriver(FullFileName, &AdapterDevHandle, RebootRequired);
+    if (ERROR_SUCCESS != Status) {
+        LogError(
+            "Error: %d. Error message: %s", Status, win32_strerror(Status).c_str());
```
petrutlucian94, line 434: ditto, let's say: "Failed to install driver. Error: .."

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+        goto exit;
+    }
+
+    if (!UpdateDriverForPlugAndPlayDevicesA(0, WNBD_HARDWAREID, FullFileName, 0, RebootRequired)) {
+        Status = GetLastError();
+        LogError(
+            "Error: %d. Error message: %s", Status, win32_strerror(Status).c_str());
```
petrutlucian94, line 441: ditto.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    DevInfoData.cbSize = sizeof DevInfoData;
+
+    if (0 == GetFullPathNameA(FileName, MAX_PATH, FullFileName, 0)) {
+        Status = GetLastError();
+        LogError(
+            "Error: %d. Error message: %s", Status, win32_strerror(Status).c_str());
+        goto exit;
+    }
+    if (FALSE == PathFileExistsA(FullFileName)) {
+        LogError("File: %s, does not exist", FullFileName);
```
petrutlucian94, line 421: nit: this looks a bit odd. How about "Could not find file: %s.".

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+        Status = ERROR_FILE_NOT_FOUND;
+        goto exit;
+    }
+    if (ERROR_SUCCESS == FindDevice(&AdapterDevHandle, &DevInfoData)) {
+        LogError("Driver already installed");
```
petrutlucian94, line 426: This is a bit odd, we're checking if there's an existing WNBD device, not an existing driver. What if there's no WNBD device but there's an existing driver?

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    CHAR FullFileName[MAX_PATH];
+    DWORD Status = ERROR_SUCCESS;
+    HDEVINFO AdapterDevHandle = INVALID_HANDLE_VALUE;
+    SP_DEVINFO_DATA DevInfoData;
+    DevInfoData.cbSize = sizeof DevInfoData;
+
+    if (0 == GetFullPathNameA(FileName, MAX_PATH, FullFileName, 0)) {
+        Status = GetLastError();
+        LogError(
+            "Error: %d. Error message: %s", Status, win32_strerror(Status).c_str());
```
petrutlucian94, line 417: How about: "Invalid file path: %s. Error..."

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+
+    DevInfoData.cbSize = sizeof DevInfoData;
+    if (!SetupDiCreateDeviceInfoA(*AdapterDevHandle, ClassName, &ClassGuid, 0, 0, DICD_GENERATE_ID, &DevInfoData)) {
+        Status = GetLastError();
+        goto exit;
+    }
+
+    if (!SetupDiSetDeviceRegistryPropertyA(*AdapterDevHandle, &DevInfoData,
+                                           SPDRP_HARDWAREID, (PBYTE)WNBD_HARDWAREID, WNBD_HARDWAREID_LEN)) {
+        Status = GetLastError();
```
petrutlucian94, line 383: Should we log all those errors? We might need to know which of those calls failed, not just InstallDriver itself. It might be useful for debugging purposes. Same applies to other errors throughout the PR.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    SP_DEVINFO_DATA DevInfoData;
+    SP_DEVINSTALL_PARAMS_A InstallParams;
+    DWORD Status = 0;
+
+    if (!SetupDiGetINFClassA(FileNameBuf, &ClassGuid, ClassName, MAX_CLASS_NAME_LEN, 0)) {
+        Status = GetLastError();
+        goto exit;
+    }
+
+    if (INVALID_HANDLE_VALUE == (*AdapterDevHandle = SetupDiCreateDeviceInfoList(&ClassGuid, 0))) {
```
petrutlucian94, line 370: It's a bit hard to tell which part of the function registers the driver and which is creating the device. Do you think it would make sense to decouple it, having InstallDriver and CreateWnbdAdapter?

libwnbd/wnbd_ioctl.cpp
```diff
@@ -133,6 +137,315 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    PWNBD_CONNECTION_LIST ConnList = NULL;
+    DWORD err = HardRemoveAllDevices();
+    if (err) {
+        return err;
+    }
+    
+    Status = WnbdRemoveDriver(AdapterDevHandle, &DevInfoData, RebootRequired);
+
+    if (INVALID_HANDLE_VALUE != AdapterDevHandle) {
+        SetupDiDestroyDeviceInfoList(AdapterDevHandle);
```
petrutlucian94, line 346: Let's catch/log this error.

petrutlucian94 (changes requested):

vstudio/wnbdevents.xml
aserdean: Documentation say it started since Vista: https://docs.microsoft.com/en-us/windows/win32/wes/eventmanifestschema-opcodetype-complextype .

The element is not mandatory and it is specified to differentiate between events which should be logged directly to the system and the ones which are logged on demand via utilities (i.e. `tracelog`).

AFAIK the default would be zero (`win:Info`) but in my opinion that would be even more confusing because that would require even more additional knowledge.

libwnbd/wnbd_ioctl.cpp
aserdean: Sure.

libwnbd/wnbd_ioctl.cpp
aserdean: Sure.

libwnbd/wnbd_ioctl.cpp
aserdean: Sure.

libwnbd/wnbd_ioctl.cpp
aserdean: Sure.

driver/wnbd_dispatch.c
```diff
@@ -182,8 +182,8 @@ NTSTATUS WnbdDispatchRequest(
                 Element->Srb->SrbStatus = SRB_STATUS_INTERNAL_ERROR;
                 CompleteRequest(Device, Element, TRUE);
                 InterlockedDecrement64(&Device->Stats.UnsubmittedIORequests);
-                WNBD_LOG_ERROR("Could not get SRB %p 0x%llx data buffer. Error: %d.",
-                               Element->Srb, Element->Tag, StorResult);
+                WNBD_LOG_WARN("Could not get SRB %p 0x%llx data buffer. Error: %d.",
```
petrutlucian94, line 185: Let's use a separate commit for the log level changes. I think there are quite a few more log messages that will have to be switched to WNBD_LOG_LOUD.

aserdean: Ooops. I've added this one by mistake.

I still need to rebase and split the PR.

libwnbd/wnbd_ioctl.cpp
petrutlucian94: > We shouldn't check if there are more than one associated driver. The purpose of the function is to fetch the driver information.

What if there are multiple drivers? What if we failed to uninstall a driver and a new one was added?

libwnbd/wnbd_ioctl.cpp
petrutlucian94: Not necessarily, our own tool might successfully install the driver but fail to create the device (e.g. due to an error).

libwnbd/wnbd_ioctl.cpp
aserdean: This function is used to retrieve the driver information data. We only return the first element found with the aid of: `AdapterDevHandle`.
That guarantees the fact that we have a valid map between the registered device associated and the driver OEM information file.

> What if there are multiple drivers? 

Disregarding the use-case of this function.

In general If there multiple drivers installed only a single one can use the hardware node `root\wnbd` (while multiple registered devices can use that node).

Our tool only tries to uninstall the driver only if there is a a registered device using the node `root\wnbd` and does a best effort mode to remove its associated driver information file.

In my opinion we should not try to make assumption and remove all of the registered devices that are under `root\wnbd`, nor should we remove all the drivers OEM information. That could be the intend of the user.

>What if we failed to uninstall a driver and a new one was added?

Again how do you install a new one?

petrutlucian94: > In my opinion we should not try to make assumption and remove all of the registered devices that are under root\wnbd, nor should we remove all the drivers OEM information. That could be the intend of the user.

I'd say removing all wnbd devices and drivers is the safest approach though, that's what "uninstall-driver" should do. If the user intends to keep some of those resources, that's an extremely unusual scenario. In that case, the user might as well just use devcon.exe exclusively.

aserdean: > I'd say removing all wnbd devices and drivers is the safest approach though, that's what "uninstall-driver" should do.

To be honest that sound more like a "clean"|"force" uninstall.
Uninstall should do the reverse of the install and nothing more.
It shouldn't automatically make assumption like restarting the computer if needed to force remove the current resources.
The user should not have additional installed drivers or devices if using `install-driver`.

petrutlucian94: > To be honest that sound more like a "clean"|"force" uninstall.
Cleaning up all WNBD resources is perfectly fine IMO.

> The user should not have additional installed drivers or devices if using install-driver.
That can happen if we fail to remove the driver before attempting to install a new version. I was hitting this quite often before using our reinstall.ps1 script if there were existing mappings.

aserdean:
This PR has been superseded.

PR #30, opened 2020-10-15
Always use an overlapped structure when calling DeviceIoControl

petrutlucian94:
This change fixes deadlocks occurring when communicating with the
WNBD driver using DeviceIoControl[1][2].

We're setting FILE_FLAG_OVERLAPPED when opening the WNBD device
in order to be able to submit multiple requests simultaneously.
The Windows docs state that in this case, an OVERLAPED structure
must be passed to functions such as ReadFile or WriteFile without
explicitly mentioning DeviceIoControl.

Apparently DeviceIoControl can hang if we don't pass an overlapped
structure, especially when having a high number of concurrent
requests. This is likely related to the event wait involved.

We're going to update the libwnbd *Ioctl* functions, adding an
optional LPOVERLAPPED argument. When not provided by the caller,
we're initalizing an internal OVERLAPED structure and wait for
the operation on the behalf of the caller, mimicking the Windows
behavior.

We're trying to re-use overlapped structures and events as much
as possible, especially in the IO path. The IO dispatch loop
will reuse the same events but we can't do much about
"WnbdSendResponse" since that will be called by libwnbd consumers
from arbitrary threads. We're adding "WnbdSendResponseEx" though,
letting the caller provide an overlapped structure.

Another advantage of adding overlapped parameters is that we allow
libwnbd consumers to perform any *Ioctl* call asynchronously.

[1] profiler counters: http://paste.openstack.org/raw/797772/
[2] wnbd / ceph counters: http://paste.openstack.org/raw/799072/

### DIFF
--- a/include/wnbd.h
+++ b/include/wnbd.h
@@ -191,6 +191,12 @@ DWORD WnbdSendResponse(
     PWNBD_IO_RESPONSE Response,
     PVOID DataBuffer,
     UINT32 DataBufferSize);
+DWORD WnbdSendResponseEx(
+    PWNBD_DISK Disk,
+    PWNBD_IO_RESPONSE Response,
+    PVOID DataBuffer,
+    UINT32 DataBufferSize,
+    LPOVERLAPPED Overlapped);
 
 /**
 * Retrieve a specific WNBD option.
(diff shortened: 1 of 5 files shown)


aserdean:

PR #32, opened 2020-10-24
Add driver properties through resource file

ml8mr:
Versioning and other details should be visible through properties on the driver file (through Windows Explorer). As Windows versions commonly use a four digit versioning scheme (x.x.x.x), use the git commit count as the fourth digit.

### DIFF
--- a/driver/wnbd.rc
+++ b/driver/wnbd.rc
@@ -24,6 +24,7 @@
 #include <windows.h>
 #include <ntverp.h>
 #include "..\include\version.h"
+#include "..\include\vendor.h"
 
 #define VER_INTERNALNAME_STR        "wnbd.sys"
 #define VER_ORIGINALFILENAME_STR    "wnbd.sys"
@@ -38,12 +39,12 @@
 #define VER_FILEVERSION_STR         WNBD_VERSION_STR_MS
 
 #undef VER_COMPANYNAME_STR
-#define VER_COMPANYNAME_STR        "SUSE LLC"
+#define VER_COMPANYNAME_STR         WNBD_COMPANY_STR
 
 #undef VER_FILEDESCRIPTION_STR
 #undef VER_PRODUCTNAME_STR
-#define VER_FILEDESCRIPTION_STR     "Windows Network Block Device"
-#define VER_PRODUCTNAME_STR         "SUSE Enterprise Storage Driver for Windows"
+#define VER_FILEDESCRIPTION_STR     WNBD_FILEDESCRIPTION_STR
+#define VER_PRODUCTNAME_STR         WNBD_PRODUCTNAME_STR
 
 #define VER_FILESUBTYPE             VFT2_DRV_SYSTEM
 
--- a/driver/wnbd.rc
+++ b/driver/wnbd.rc
@@ -1,0 +1,58 @@
+/*-
+ * SPDX-License-Identifier: LGPL-2.1-only
+ *
+ * Copyright (C) 2019-2020 SUSE LLC
+ *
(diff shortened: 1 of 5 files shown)


driver/wnbd.rc
```diff
@@ -0,0 +1,58 @@
...
+#define VER_PRODUCTVERSION          WNBD_VERSION_MAJOR,WNBD_VERSION_MINOR,WNBD_VERSION_PATCH,WNBD_COMMIT_COUNT
+
+#undef VER_PRODUCTVERSION_STR
+#define VER_PRODUCTVERSION_STR      WNBD_VERSION_STR
+
+#define VER_FILEVERSION             WNBD_VERSION_MAJOR,WNBD_VERSION_MINOR,WNBD_VERSION_PATCH,WNBD_COMMIT_COUNT
+#define VER_FILEVERSION_STR         WNBD_VERSION_STR_MS
+
+#undef VER_COMPANYNAME_STR
+#define VER_COMPANYNAME_STR        "SUSE LLC"
```
alexpilotti, line 41: All branding should go in variables ideally

driver/wnbd.rc
```diff
@@ -0,0 +1,58 @@
...
+
+#undef VER_FILEDESCRIPTION_STR
+#undef VER_PRODUCTNAME_STR
+#define VER_FILEDESCRIPTION_STR     "Windows Network Block Device"
+#define VER_PRODUCTNAME_STR         "SUSE Enterprise Storage Driver for Windows"
```
alexpilotti, line 46: ditto

driver/wnbd.rc
```diff
@@ -0,0 +1,58 @@
...
+
+#define VER_FILESUBTYPE             VFT2_DRV_SYSTEM
+
+#define VER_FILETYPE                VFT_DRV
+
+#define VER_LEGALCOPYRIGHT_STR      "Copyright \251 2019-2020 SUSE LLC All rights reserved.", "\0"
```
alexpilotti, line 52: This one IMO is ok at the current stage, not sure once we move it upstream (LF).

driver/wnbd.rc
ml8mr: Through something like an include/vendor.h, or did you have something else in mind?

alexpilotti: A common way to do this is to directly assign #DEFINE from project property variables at build time, without generating header files, see for example:

https://github.com/cloudbase/cloudbase-init-installer/blob/master/BuildAutomation/BuildCloudbaseInitSetup.ps1#L154

Those project properties are defined in the project's property pages.

alexpilotti: Update: please disregard the last comment, beside automating the build we also want to ensure that people can just open the solution in VS and hit build, so having _version.h_ and _vendor.h_ files is the easiest approach rather than using variables.

What about adding a vendor.h file to the project with empty strings and a #warning to let people know they have to set those #defines?

ml8mr: How about a vendor.h that uses a conditional (set by DefineConstants) to determine which vendor properties to apply? We can default to upstream (Cloudbase) strings, and easily allow extension by other vendors.

I've got an additional commit which shows this, and seems to work here. I'll push it shortly.

alexpilotti: What about leaving a generic "WNBD" for all three strings? The idea would be to overwrite Vendor.h during the build, this way we don't have to include all the branding info in the project itself.

ml8mr: Ok, we can go that route for now.

PR #33, opened 2020-10-26
driver: clean up IO queue when resetting adapter

petrutlucian94:
When hitting timeouts, Storport resets the adapter. NBD replies
don't include the IO size so we keep that information around
when resetting the adapter. We don't need the payload itself,
just the size so that we can tell where the next NBD header
starts.

We don't have to do this for the IOCTL based IO channel, in
which case we're going to clean up the IO queue.


aserdean:

PR #36, opened 2020-10-28
libwnbd: pin structure sizes

petrutlucian94:
We're going to add structure size assertions to ensure that we aren't
breaking backwards compatibility by mistake.

This is also helps spotting portability issues as other compilers
might use different structure layouts, especially when having
bitfields.


aserdean:

PR #37, opened 2020-10-30
Add version detection fallback

petrutlucian94:
When building the project, we're automatically detecting the WNBD
version using the git tag.

The issue is that we shouldn't rely on it, users might just download
a zip of the archive.

This change will generate a version.h template that the user can fill
when the version detection fails.


aserdean:

PR #38, opened 2020-11-03
Add command line options to install and uninstall the driver

aserdean:
This patch introduces two more supported CLI commands.
`install-driver`, `clean-uninstall`.

The hardware ID and length are predefined, the only required installation
argument is the INF file location.

The clean uninstall sequence will remove any created devices and it will
also try to remove any previous driver installations by  cycling through
all the OEM driver installations and trying to find a match over
the catalog file `wnbd.cat`.

Invoking the clean uninstall will also hard disconnect any online mappings.

Document the usage of the client utility `wnbd-client` to install
and clean uninstall the driver.

### DIFF
--- a/libwnbd/wnbd_ioctl.cpp
+++ b/libwnbd/wnbd_ioctl.cpp
@@ -8,15 +8,20 @@
 #include <windef.h>
 #include <winioctl.h>
 #include <winioctl.h>
+#include <newdev.h>
 #include <ntddscsi.h>
+#include "Shlwapi.h"
 #include <setupapi.h>
 #include <string.h>
+#include <infstr.h>
 
 #include "wnbd.h"
 #include "wnbd_log.h"
 #include "utils.h"
 
 #pragma comment(lib, "Setupapi.lib")
+#pragma comment(lib, "Newdev.lib")
+#pragma comment(lib, "Shlwapi.lib")
 
 #define STRING_OVERFLOWS(Str, MaxLen) (strlen(Str + 1) > MaxLen)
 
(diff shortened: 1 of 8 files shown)


libwnbd/wnbd_ioctl.cpp
```diff
@@ -139,6 +144,367 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    HANDLE DirHandle = INVALID_HANDLE_VALUE;
+    WIN32_FIND_DATA OemFindData;
+
+    if (!GetWindowsDirectory(OemName, ARRAYSIZE(OemName))) {
+        Status = GetLastError();
+        LogError("GetWindowsDirectory failed with "
+            "error: %d. Error message: %s", Status, win32_strerror(Status).c_str());
+        return;
+    }
+    if (wcscat_s(OemName, ARRAYSIZE(OemName), L"\\INF\\OEM*.INF")) {
```
petrutlucian94, line 210: Instead of parsing all *.inf files, isn't it better to just use the ``SetupDi`` API to retrieve the right one, like you did before? https://github.com/cloudbase/wnbd/commit/57f523285da54f8963c4f5f7a5048466a1803866#diff-0a246dbc425e428f1d11ae42c36d302a5fe6ed52d3c3a3e499a49e565752256dR140

libwnbd/wnbd_ioctl.cpp
```diff
@@ -139,6 +144,367 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    Status = WnbdList(ConnList, &BufferSize);
+    if (Status) {
+        goto exit;
+    }
+    CHAR* InstanceName;
+    for (ULONG index = 0; index < ConnList->Count; index++) {
+        InstanceName = ConnList->Connections[index].Properties.InstanceName;
+        WNBD_REMOVE_OPTIONS RemoveOptions = { 0 };
+        RemoveOptions.Flags.HardRemoveFallback = TRUE;
+        RemoveOptions.SoftRemoveTimeoutMs = 30 * 1000;
```
petrutlucian94, line 327: This can be an issue: soft removing all disks sequentially with a 30s timeout. For just 20 in-use volumes, you might have to wait 10 minutes.

FWIW, in ceph I'm using a thread pool and a global timeout: https://github.com/petrutlucian94/ceph/blob/windows.12/src/tools/rbd_wnbd/rbd_wnbd.cc#L462-L523

I think we can move that ceph code to libwnbd, maybe adding WnbdRemoveAll with an optional OwnerName argument. Until then, it's probably better to stick to a force removal.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -139,6 +144,367 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+        Status = GetLastError();
+        LogError(
+            "Could not remove driver. "
+            "Error: %d. Error message: %s", Status, win32_strerror(Status).c_str());
+    }
+
+    return Status;
+}
+
+// We assume we are using a single WNBD adapter per host
```
petrutlucian94, line 250: This is used as part of a loop, removing all WNBD devices, which is great, but it contradicts this inline comment stating that we assume we're only using one adapter per host. I guess we can remove the comment.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -139,6 +144,367 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+            if (!SetupDiDestroyDeviceInfoList(AdapterDevHandle)) {
+                Status = GetLastError();
+                LogError("SetupDiDestroyDeviceInfoList failed with "
+                    "error: %d. Error message: %s", Status, win32_strerror(Status).c_str());
+            }
+        }
+
+        if (ERROR_SUCCESS != Status) {
+            LogError(
+                "Error: %d. Error message: %s", Status, win32_strerror(Status).c_str());
```
petrutlucian94, line 374: An error without any context isn't very useful. Plus, it's redundant since all the above errors are logged so we can just remove it.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -139,6 +144,367 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+
+    while (Status == ERROR_SUCCESS) {
+        Status = FindWnbdAdapterDevice(&AdapterDevHandle, &DevInfoData);
+        if (ERROR_SUCCESS != Status) {
+            goto exit;
+        }
+
+        Status = RemoveWnbdDevice(AdapterDevHandle, &DevInfoData, RebootRequired);
+
+        if (INVALID_HANDLE_VALUE != AdapterDevHandle) {
```
petrutlucian94, line 364: Do we need this check? Wouldn't FindWnbdAdapterDevice return an error?

wnbd-client/client.cpp
```diff
@@ -352,4 +378,15 @@ Client::Command commands[] = {
...
         execute_reset_opt, get_reset_opt_args),
+    Client::Command(
+        "install-driver", {}, "Install WNBD driver and create its adapter",
+        execute_install, get_filename_args),
+    Client::Command(
+        "clean-uninstall", {}, "Remove all disk mappings with a SOFT disconnect option"
+                               " and tries to uninstall all installed WNBD drivers.",
+        execute_clean_uninstall),
+    Client::Command(
+        "force-clean-uninstall", {}, "Remove all disk mappings with a HARD disconnect option"
```
petrutlucian94, line 389: Why not just add a "hard-disconnect" parameter, matching the "unmap" one? The timeout might become configurable as well.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -139,6 +144,367 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+        Status = RemoveWnbdInf(OemFindData.cFileName);
+        if (Status) {
+            LogError("Failed while trying to remove OEM file: %ls",
+                OemFindData.cFileName);
+        }
+    } while (FindNextFile(DirHandle, &OemFindData));
+
+    FindClose(DirHandle);
+
+    return;
```
petrutlucian94, line 230: Why not propagate the errors?

petrutlucian94 (changes requested):

libwnbd/wnbd_ioctl.cpp
aserdean: That would be wonderful :), but unfortunately `SetupDi` works only on registered devices and not the driver catalog. Same applies for the `Di` APIs.

From what I found out it is either parsing all the OEM*.inf files or cycling through all the driver registry entries.

libwnbd/wnbd_ioctl.cpp
aserdean: It's a best effort to remove the drivers from the driver catalog store.
We may end up in the situation in which the current driver needs a reboot to remove the device.
The inf file can't be removed until all the driver devices are removed, but we want all other drivers.
I don't see any clear advantage to propagate the error beside just logging what failed in this case.

petrutlucian94: > We may end up in the situation in which the current driver needs a reboot to remove the device.

Would it still need a reboot, considering that we're removing all devices?

> I don't see any clear advantage to propagate the error beside just logging what failed in this case.

When the uninstall command returns 0, one might expect all WNBD devices and drivers to be gone.

libwnbd/wnbd_ioctl.cpp
aserdean: Calling SetupDiDestroyDeviceInfoList on `INVALID_HANDLE_VALUE` is undefined AFAIR, but the documentation doesn't mention it so will remove it.

libwnbd/wnbd_ioctl.cpp
aserdean: Thanks for pointing that out. I added a TODO for it.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -139,6 +144,358 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+        return ERROR_BAD_PATHNAME;
+    }
+    DirHandle = FindFirstFile(OemName, &OemFindData);
+    if (DirHandle == INVALID_HANDLE_VALUE) {
+        /* No OEM files */
+        return Status;
+    }
+
+    do {
+        Status = RemoveWnbdInf(OemFindData.cFileName);
```
petrutlucian94, line 222: We should preserve the first or the last non-zero status. If out of 5 files we fail to remove one, the status should be non-zero.

libwnbd/wnbd_ioctl.cpp
```diff
@@ -139,6 +144,358 @@ DWORD WnbdOpenAdapterCMDeviceInstance(PDEVINST DeviceInstance)
...
+    }
+    Status = WnbdList(ConnList, &BufferSize);
+    if (Status) {
+        goto exit;
+    }
+    CHAR* InstanceName;
+    for (ULONG index = 0; index < ConnList->Count; index++) {
+        InstanceName = ConnList->Connections[index].Properties.InstanceName;
+        WNBD_REMOVE_OPTIONS RemoveOptions = { 0 };
+        /* TODO add parallel and soft disconnect removal */
```
petrutlucian94, line 326: Once we do that, we can probably move it to libwnbd.cpp/wnbd.h so that library consumers can use it as well. I wouldn't do it now though since we want those additional arguments

wnbd-client/client.cpp
```diff
@@ -352,4 +373,11 @@ Client::Command commands[] = {
     Client::Command(
         "reset-opt", {}, "Reset driver option.",
         execute_reset_opt, get_reset_opt_args),
+    Client::Command(
+        "install-driver", {}, "Install WNBD driver and create its adapter",
+        execute_install, get_filename_args),
+    Client::Command(
+        "clean-uninstall", {}, "Remove all disk mappings with a HARD disconnect option"
```
petrutlucian94, line 380: nit: I'm not a big fan of "clean-uninstall", wondering if we should stick to "uninstall-driver". It's quite obvious that the existing mappings might be removed when the driver is uninstalled.

Also, "Remove all disk mappings with a HARD disconnect option" sounds a bit weird, how about "Hard remove all disk mappings and adapters and uninstall all WNBD driver instances".

petrutlucian94 (changes requested):
Almost there.

libwnbd/wnbd_ioctl.cpp
petrutlucian94: My bad, I was expecting it to loop over all drivers and return an error afterwards. Now that I see that we're breaking the loop when hitting an error, I guess it's fine.

wnbd-client/client.cpp
```diff
@@ -316,6 +316,27 @@ DWORD execute_reset_opt(const po::variables_map& vm)
         safe_get_param<bool>(vm, "persistent"));
 }
 
+DWORD execute_install(const po::variables_map& vm)
+{
+    return CmdInstall(
+        safe_get_param<string>(vm, "filename").c_str());
+}
+
+void get_filename_args(
```
petrutlucian94, line 325: Nit: let's rename this to "get_install_args". We might add other parameters, plus the filename description is too tied to this operation to be used in other scenarios.

wnbd-client/client.cpp
```diff
@@ -316,6 +316,27 @@ DWORD execute_reset_opt(const po::variables_map& vm)
...
+    po::positional_options_description& positonal_opts,
+    po::options_description& named_opts)
+{
+    positonal_opts.add("filename", 1);
+    named_opts.add_options()
+        ("filename", po::value<string>()->required(), "Absolute or relative path to"
```
petrutlucian94, line 331: Nit: probably it's enough to say something like "The OEM driver information (wnbd.inf) path".

petrutlucian94 (approved):

PR #39, opened 2020-11-04
Clean log messages

aserdean:
This PR is intended to clean the wnbd log messages.

- Remove leading `: ` from log messages.

- Fix format specifiers.

- Demote some log messages.

- Print NTSTATUS as hex instead of integer format.

- Remove extraneous arguments passed to log messages

### DIFF
--- a/driver/driver.c
+++ b/driver/driver.c
@@ -119,8 +119,8 @@ WnbdDispatchPnp(PDEVICE_OBJECT DeviceObject,
     ASSERT(IoLocation);
     UCHAR MinorFunction = IoLocation->MinorFunction;
 
-    WNBD_LOG_LOUD("Received PnP request: %s (%d).",
-                  WnbdToStringPnpMinorFunction(MinorFunction), MinorFunction);
+    WNBD_LOG_DEBUG("Received PnP request: %s (%d).",
+                   WnbdToStringPnpMinorFunction(MinorFunction), MinorFunction);
     switch (MinorFunction) {
     case IRP_MN_QUERY_CAPABILITIES:
         IoLocation->Parameters.DeviceCapabilities.Capabilities->SilentInstall = 1;
@@ -161,7 +161,7 @@ WnbdDispatchPnp(PDEVICE_OBJECT DeviceObject,
 
     Status = StorPortDispatchPnp(DeviceObject, Irp);
 
-    WNBD_LOG_LOUD("Exit: 0x%x", Status);
+    WNBD_LOG_DEBUG("Exit: 0x%x", Status);
     return Status;
 }
 
--- a/driver/driver.c
+++ b/driver/driver.c
@@ -29,8 +29,6 @@ NTSTATUS
 DriverEntry(PDRIVER_OBJECT DriverObject,
             PUNICODE_STRING RegistryPath)
 {
-    WNBD_LOG_LOUD(": Enter");
-
     /*
      * Register Virtual Storport Miniport data
      */
@@ -103,8 +101,6 @@ DriverEntry(PDRIVER_OBJECT DriverObject,
     DriverObject->MajorFunction[IRP_MJ_PNP] = 0 != StorPortDispatchPnp ? WnbdDispatchPnp : 0;
     GlobalExt = NULL;
 
-    WNBD_LOG_LOUD(": Exit");
-
     /*
      * Report status in upper layers
      */
@@ -116,7 +112,6 @@ NTSTATUS
 WnbdDispatchPnp(PDEVICE_OBJECT DeviceObject,
                 PIRP Irp)
 {
-    WNBD_LOG_LOUD(": Enter");
     ASSERT(Irp);
     NTSTATUS Status = STATUS_INVALID_DEVICE_REQUEST;
     PIO_STACK_LOCATION IoLocation = IoGetCurrentIrpStackLocation(Irp);
@@ -146,7 +141,7 @@ WnbdDispatchPnp(PDEVICE_OBJECT DeviceObject,
             }
             Status = WnbdGetScsiAddress(DeviceObject, &ScsiAddress);
             if (Status) {
-                WNBD_LOG_ERROR("Could not query SCSI address. Error: %d.", Status);
+                WNBD_LOG_WARN("Could not query SCSI address. Error: 0x%x.", Status);
                 break;
             }
 
@@ -166,7 +161,7 @@ WnbdDispatchPnp(PDEVICE_OBJECT DeviceObject,
 
     Status = StorPortDispatchPnp(DeviceObject, Irp);
 
-    WNBD_LOG_LOUD(": Exit: %d", Status);
+    WNBD_LOG_LOUD("Exit: 0x%x", Status);
     return Status;
 }
 
@@ -174,11 +169,8 @@ _Use_decl_annotations_
 VOID
 WnbdDriverUnload(PDRIVER_OBJECT DriverObject)
 {
-    WNBD_LOG_LOUD(": Enter");
 
     if (0 != StorPortDriverUnload) {
         StorPortDriverUnload(DriverObject);
     }
-
-    WNBD_LOG_LOUD(": Exit");
 }
--- a/driver/nbd_dispatch.c
+++ b/driver/nbd_dispatch.c
@@ -111,9 +111,9 @@ WnbdProcessDeviceThreadRequests(_In_ PWNBD_DISK_DEVICE Device)
             ExInterlockedInsertTailList(
                 &Device->SubmittedReqListHead,
                 &Element->Link, &Device->SubmittedReqListLock);
-            WNBD_LOG_LOUD("Sending %s request. Address: %p Tag: 0x%llx. FUA: %d",
-                          NbdRequestTypeStr(NbdReqType), Element->Srb, Element->Tag,
-                          Element->FUA);
+            WNBD_LOG_DEBUG("Sending %s request. Address: %p Tag: 0x%llx. FUA: %d",
+                           NbdRequestTypeStr(NbdReqType), Element->Srb, Element->Tag,
+                           Element->FUA);
 
             if(NbdReqType == NBD_CMD_WRITE){
                 Status = WnbdRequestWrite(Device, Element,
@@ -236,8 +236,8 @@ NbdProcessDeviceThreadReplies(_In_ PWNBD_DISK_DEVICE Device)
         // We need to avoid accessing aborted or already completed SRBs.
         PCDB Cdb = (PCDB)&Element->Srb->Cdb;
         int NbdReqType = ScsiOpToNbdReqType(Cdb->AsByte[0]);
-        WNBD_LOG_LOUD("Received reply header for %s %p 0x%llx.",
-                      NbdRequestTypeStr(NbdReqType), Element->Srb, Element->Tag);
+        WNBD_LOG_DEBUG("Received reply header for %s %p 0x%llx.",
+                       NbdRequestTypeStr(NbdReqType), Element->Srb, Element->Tag);
 
         if(IsReadSrb(Element->Srb)) {
             StorResult = StorPortGetSystemAddress(Element->DeviceExtension, Element->Srb, &SrbBuff);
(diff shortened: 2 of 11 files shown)


driver/driver.c
```diff
@@ -89,7 +89,7 @@ DriverEntry(PDRIVER_OBJECT DriverObject,
                                  (PHW_INITIALIZATION_DATA)&WnbdInitData,
                                  NULL);
     if (!NT_SUCCESS(Status)) {
-        WNBD_LOG_ERROR("DriverEntry failure in call to StorPortInitialize. Status: 0x%x", Status);
+        WNBD_LOG_ERROR("DriverEntry failure in call to StorPortInitialize. Status: %x", Status);
```
petrutlucian94, line 92: Let's not use "%x", we can stick to "0x%x" or "%#x". Without the "0x" prefix, it's hard to tell if that's a decimal or hexadecimal number if we get something like "Status: 10".

driver/driver.c
```diff
@@ -29,7 +29,7 @@ NTSTATUS
 DriverEntry(PDRIVER_OBJECT DriverObject,
             PUNICODE_STRING RegistryPath)
 {
-    WNBD_LOG_LOUD(": Enter");
+    WNBD_LOG_LOUD("Enter");
```
petrutlucian94, line 32: I'm wondering if we should keep those ``` WNBD_LOG_LOUD("Enter");```

driver/driver.c
```diff
@@ -103,8 +103,7 @@ DriverEntry(PDRIVER_OBJECT DriverObject,
     DriverObject->MajorFunction[IRP_MJ_PNP] = 0 != StorPortDispatchPnp ? WnbdDispatchPnp : 0;
     GlobalExt = NULL;
 
-    WNBD_LOG_LOUD(": Exit");
-
+    WNBD_LOG_LOUD("Exit");
```
petrutlucian94, line 106: Using one commit per file is weird and pollutes the git log, let's squash them.

driver/nbd_dispatch.c
```diff
@@ -234,7 +240,7 @@ NbdProcessDeviceThreadReplies(_In_ PWNBD_DISK_DEVICE Device)
     }
     KeReleaseSpinLock(&Device->SubmittedReqListLock, Irql);
     if(!Element) {
-        WNBD_LOG_ERROR("Received reply with no matching request tag: 0x%llx",
+        WNBD_LOG_WARN("Received reply with no matching request tag: 0x%llx",
```
petrutlucian94, line 243: Let's use a debug message here. This will be printed for every abandoned request and there can be lots of them.

driver/nbd_protocol.c
```diff
@@ -187,16 +189,16 @@ NbdSendOptExportName(_In_ INT Fd,
...
 
     NTSTATUS error = STATUS_SUCCESS;
 
     NbdSendRequest(Fd, NBD_OPT_EXPORT_NAME, strlen(Name), Name);
     CHAR Buf[sizeof(*Flags) + sizeof(*Size)];
     RtlZeroMemory(Buf, sizeof(*Flags) + sizeof(*Size));
     if (NbdReadExact(Fd, Buf, sizeof(Buf), &error) < 0 && Go) {
-        WNBD_LOG_ERROR("Server does not support NBD_OPT_GO and"
-            "dropped connection after sending NBD_OPT_EXPORT_NAME.");
+        WNBD_LOG_INFO("Server does not support NBD_OPT_GO and"
```
petrutlucian94, line 200: This should be a warning, at least.

driver/nbd_protocol.c
```diff
@@ -228,14 +230,14 @@ NbdNegotiate(_In_ INT* Pfd,
     RtlZeroMemory(Buf, 8);
     NbdReadExact(Fd, Buf, 8, &status);
     if (strcmp(Buf, INIT_PASSWD)) {
-        WNBD_LOG_ERROR("INIT_PASSWD");
+        WNBD_LOG_INFO("INIT_PASSWD");
```
petrutlucian94, line 233: Let's also update the message, maybe something like "Received NBD INIT_PASSWD". Also, I guess it can be a debug message.

driver/nbd_protocol.c
```diff
@@ -139,16 +140,17 @@ NbdReadHandshakeReply(_In_ INT Fd)
     Retval->Datasize = RtlUlongByteSwap(Retval->Datasize);
 
     if (REPLY_MAGIC != Retval->Magic) {
-        WNBD_LOG_ERROR("Received invalid negotiation magic %llu (expected %llu)",
-                       Retval->Magic, REPLY_MAGIC);
+        WNBD_LOG_WARN("Received invalid negotiation magic %llu (expected %llu)",
```
petrutlucian94, line 143: I'm wondering if we should stick to WNBD_LOG_ERROR. It's a fatal error, meaning that the NBD connection cannot be established.

driver/scsi_function.c
```diff
@@ -133,7 +133,7 @@ WnbdExecuteScsiFunction(PVOID DeviceExtension,
     if (SrbGetCdb(Srb)) {
         BYTE CdbValue = SrbGetCdb(Srb)->AsByte[0];
 
-        WNBD_LOG_INFO(": Received %#02x command. SRB = 0x%p. CDB = 0x%x. PathId: %d TargetId: %d LUN: %d",
+        WNBD_LOG_INFO("Received %#02x command. SRB = 0x%p. CDB = 0x%x. PathId: %d TargetId: %d LUN: %d",
```
petrutlucian94, line 136: Let's use a debug message instead.

driver/scsi_operation.c
```diff
@@ -629,13 +629,13 @@ WnbdPendOperation(_In_ PWNBD_EXTENSION DeviceExtension,
         }
         break;
     default:
-        WNBD_LOG_ERROR("Unknown Pending SCSI Operation received");
+        WNBD_LOG_INFO("Unknown Pending SCSI Operation received");
```
petrutlucian94, line 632: Let's update the message to WNBD_LOG_INFO("Unknown SCSI operation: %#x", Cdb->AsByte[0])

driver/userspace.c
```diff
@@ -374,21 +374,21 @@ WnbdCreateConnection(PWNBD_EXTENSION DeviceExtension,
     if (!Device->Properties.BlockSize || !Device->Properties.BlockCount ||
         Device->Properties.BlockCount > ULLONG_MAX / Device->Properties.BlockSize)
     {
-        WNBD_LOG_ERROR("Invalid block size or block count. "
-                       "Block size: %d. Block count: %lld.",
-                       Device->Properties.BlockSize,
-                       Device->Properties.BlockCount);
+        WNBD_LOG_WARN("Invalid block size or block count. "
```
petrutlucian94, line 377: Let's stick to an error, we failed to create the connection because of that.

Probably a good rule of thumb: if we manage to resume the connection after the error we can use a warning, if not, we should probably use error level.

driver/userspace.c
```diff
@@ -716,20 +716,20 @@ WnbdParseUserIOCTL(PWNBD_EXTENSION DeviceExtension,
...
         if (!strlen((PSTR) &StatsCmd->InstanceName)) {
-            WNBD_LOG_ERROR("WNBD_STATS: Invalid instance name");
+            WNBD_LOG_WARN("WNBD_STATS: Invalid instance name");
             Status = STATUS_INVALID_PARAMETER;
             break;
         }
 
         if (!Irp->AssociatedIrp.SystemBuffer ||
-                CHECK_O_LOCATION(IoLocation, WNBD_DRV_STATS)) {
+            CHECK_O_LOCATION(IoLocation, WNBD_DRV_STATS)) {
```
petrutlucian94, line 732: Unnecessary change, now it's more difficult to see where the if statement ends.

driver/userspace.c
```diff
@@ -650,7 +650,7 @@ WnbdParseUserIOCTL(PWNBD_EXTENSION DeviceExtension,
         if (!Irp->AssociatedIrp.SystemBuffer ||
             CHECK_O_LOCATION_SZ(IoLocation, RequiredBuffSize))
         {
-            WNBD_LOG_ERROR("IOCTL_WNBD_LIST: Bad output buffer");
+            WNBD_LOG_WARN("IOCTL_WNBD_LIST: Bad output buffer");
```
petrutlucian94, line 653: Let's use a debug message here. Each list command will get a warning since we're adjusting the buffer size.

driver/userspace.c
```diff
@@ -908,7 +908,7 @@ WnbdParseUserIOCTL(PWNBD_EXTENSION DeviceExtension,
             (PWNBD_IOCTL_LIST_DRV_OPT_COMMAND) Irp->AssociatedIrp.SystemBuffer;
 
         if (!ListOptCmd || CHECK_I_LOCATION(IoLocation, IOCTL_WNBD_LIST_DRV_OPT)) {
-            WNBD_LOG_ERROR("IOCTL_WNBD_RESET_DRV_OPT: Bad input or output buffer");
+            WNBD_LOG_WARN("IOCTL_WNBD_RESET_DRV_OPT: Bad input or output buffer");
```
petrutlucian94, line 911: Let's use a debug message here. Each list command will get a warning since we're adjusting the buffer size.

driver/wnbd_dispatch.c
```diff
@@ -297,15 +297,15 @@ NTSTATUS WnbdHandleResponse(
...
+                WNBD_LOG_WARN("Could not get SRB %p 0x%llx data buffer. Error: %lu.",
+                              Element->Srb, Element->Tag, StorResult);
                 Element->Srb->SrbStatus = SRB_STATUS_INTERNAL_ERROR;
                 Status = STATUS_INTERNAL_ERROR;
                 goto Exit;
             }
         }
     } else {
-        WNBD_LOG_WARN("Received reply header for aborted request: %p 0x%llx.",
+        WNBD_LOG_INFO("Received reply header for aborted request: %p 0x%llx.",
```
petrutlucian94, line 308: Ditto, let's use a debug message.

petrutlucian94 (changes requested):

driver/driver.c
aserdean: It's easier to see the changes in my opinion, but sure I will squash them in one commit.

driver/driver.c
aserdean: It will help in the event of stack corruption (very small chance for it to happen, please let me know if you want to keep them).

driver/driver.c
aserdean: Sure, I will change the rest of the occurrences.

driver/userspace.c
aserdean: I will revert the change, although both seem terrible. I'm wondering if we should move the logical operator on a new line, that will help with the visibility.

driver/nbd_dispatch.c
```diff
@@ -184,12 +177,12 @@ NbdDeviceRequestThread(_In_ PVOID Context)
     }
 
     PsTerminateSystemThread(STATUS_SUCCESS);
+    return;
```
petrutlucian94, line 180: Unrelated change, probably mistake.

driver/nbd_dispatch.c
```diff
@@ -201,12 +194,12 @@ NbdDeviceReplyThread(_In_ PVOID Context)
...
```
petrutlucian94, line 197: ditto

driver/nbd_protocol.c
```diff
@@ -392,7 +373,7 @@ NbdOpenAndConnect(PCHAR HostName,
     if (Ai) {
         FreeAddrInfo(Ai);
     }
-    WNBD_LOG_LOUD(": Exit");
+    WNBD_LOG_LOUD("Exit");
```
petrutlucian94, line 376: Leftover

driver/nbd_protocol.c
```diff
@@ -433,7 +413,7 @@ NbdRequest(
     }
 Exit:
     *IoStatus = Status;
-    WNBD_LOG_LOUD(": Exit");
+    WNBD_LOG_LOUD("Exit");
```
petrutlucian94, line 416: Ditto.

driver/userspace.c
petrutlucian94: Yep, that would work as well.

driver/nbd_dispatch.c
petrutlucian94: Ughh, my bad. We must use WNBD_LOG_ERROR for nbd_dispatch since that will close the NBD connection. It's a non-fatal error for wnbd_dispatch, where I see that you're already using WNBD_LOG_TRACE (might be worth using WNBD_LOG_DEBUG for consistency).

petrutlucian94 (approved):
Looks good, let's wait for the tests to complete and then we can merge it. Thanks for addressing the comments!

PR #40, opened 2020-11-04
Add etw support

aserdean:
Add ETW support for the kernelspace.

For now, we define and use a single variable event which reuses
the messages relayed via WnbdLog.

To install and remove the custom events we need to use the utility:
`wevtutil`

To start a trace session one can use:
tracelog -start WNBDEventdrv -guid #FFACC4E7-C115-4FE2-9D3C-80FAE73BAB91 -f WNBDEventdrv.etl
To stop:
tracelog -stop WNBDEventdrv
To display the trace use:
tracerpt WNBDEventdrv.etl

WNBD error and warning messages are automatically displayed inside the default `SYSTEM` channel.

### DIFF
--- a/driver/debug.c
+++ b/driver/debug.c
@@ -10,6 +10,7 @@
 
 #include "debug.h"
 #include "options.h"
+#include "events.h"
 
 #define WNBD_LOG_BUFFER_SIZE 512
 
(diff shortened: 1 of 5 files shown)


petrutlucian94 (approved):
Looks good.

PR #41, opened 2020-11-04
Retrive disk number and pnp id at driver level

petrutlucian94:
Retrive disk number and pnp id at driver level

At the moment, libwnbd retrieves the disk number and PNP id using
a WMI query per disk, which can be extremely inefficient when having
a large amount of disks.

This change will fetch it at the driver level instead.


libwnbd/libwnbd.vcxproj
```diff
@@ -19,14 +19,12 @@
     <ClCompile Include="utils.cpp" />
     <ClCompile Include="wnbd_ioctl.cpp" />
     <ClCompile Include="wnbd_log.c" />
```
aserdean, line 21: wnbd_wmi.cpp[h] needs to be remove from libwnbd/libwnbd.vcxproj.filters as well

driver/driver.c
```diff
@@ -130,6 +130,52 @@ WnbdDispatchPnp(PDEVICE_OBJECT DeviceObject,
...
+                                  Status);
+                }
+            }
+            else {
+                WNBD_LOG_WARN("Couldn't not get disk number. "
+                              "Couldn't get attached PDO.");
+            }
+            ObDereferenceObject(AttachedDisk);
+
+            DWORD RequiredSize = 0;
```
aserdean, line 167: Consider moving `RequiredSize` to WnbdGetDiskInstancePath as you did for the type: https://github.com/cloudbase/wnbd/pull/41/files#diff-580945d31a61840d923a5b5a149d0a0e645d44898b80d8eed25ea0713b208e00R468.

aserdean (changes requested):

driver/driver.c
petrutlucian94: You mean the ``ReturnedType``? I was thinking about letting the caller adjust the buffer size, if needed.

In this case, we're using WNBD_MAX_NAME_LENGTH(255), which is greater than MAX_DEVICE_ID_LEN(200), so we should be fine.

libwnbd/libwnbd.vcxproj
petrutlucian94: Good point, thanks!

aserdean (approved):

PR #66, opened 2022-06-08
Fix Appveyor job

petrutlucian94:
Fix Appveyor job

The CI job is currently failing as the signing algorithm is set for "Release" and "Debug" builds but not for the "Analyze" one.

At the same time, the Analyze job is a resource intensive operation which now times out under Appveyor, for which reason we're going to temporarily disable it.

When preparing or reviewing driver changes, we recommend running the static code analysis by invoking the "Analyze" build locally.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,9 @@ branches:
   only:
   - master
 configuration:
-  - Analyze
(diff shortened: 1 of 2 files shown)


ionutbalutoiu (approved):
LGTM

PR #70, opened 2022-09-28
Add persistent reservation support

petrutlucian94:
Add persistent reservation support

We're making the necessary driver and libwnbd changes so that userspace daemons may receive scsi persistent reservation requests, when advertised as supported.

To keep things simple, we're only defining two libwnbd callbacks: ``PersistInFunc`` and ``PersistOutFunc``. The libwnbd consumers will have to decode the scsi operation parameters and return the expected data.

### DIFF
--- a/driver/scsi_operation.c
+++ b/driver/scsi_operation.c
@@ -605,6 +605,22 @@ WnbdPendOperation(_In_ PWNBD_EXTENSION DeviceExtension,
             FALSE);
         }
         break;
+    case SCSIOP_PERSISTENT_RESERVE_IN:
+    case SCSIOP_PERSISTENT_RESERVE_OUT:
+        ULONG DataTransferLength = SrbGetDataTransferLength(Srb);
+
+        if (DataTransferLength > WNBD_DEFAULT_MAX_TRANSFER_LENGTH)
+        {
+            Srb->SrbStatus = SRB_STATUS_ABORTED;
+            Status = STATUS_BUFFER_TOO_SMALL;
+            break;
+        }
+
+        Status = WnbdPendElement(DeviceExtension, Device, Srb,
+            0,
+            DataTransferLength,
+            FALSE);
+        break;
     default:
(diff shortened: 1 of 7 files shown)


driver/scsi_operation.c
```diff
@@ -605,6 +605,24 @@ WnbdPendOperation(_In_ PWNBD_EXTENSION DeviceExtension,
...
+        ULONG DataTransferLength = SrbGetDataTransferLength(Srb);
+
+        if (DataTransferLength > WNBD_DEFAULT_MAX_TRANSFER_LENGTH)
+        {
+            Srb->SrbStatus = SRB_STATUS_ABORTED;
+            Status = STATUS_BUFFER_TOO_SMALL;
+            break;
+        }
+
+        // TODO: which size do we specify here? Is it the allocation
```
petrutlucian94, line 619: TODO: keep data transfer length but ensure that the allocation length doesn't exceed it.

driver/util.c
```diff
@@ -327,20 +338,37 @@ ValidateScsiRequest(
                 "Write, flush or trim requested on a read-only disk.");
             return FALSE;
         }
+    // TODO: should we allow changing persistent reservations for
```
petrutlucian94, line 341: TODO: let's reject PR OUT for RO disks.

driver/wnbd_dispatch.c
```diff
@@ -218,6 +196,53 @@ NTSTATUS WnbdDispatchRequest(
...
+            REVERSE_BYTES_2(
+                &Request->Cmd.PersistResOut.ParameterListLength,
+                &Cdb->PERSISTENT_RESERVE_OUT.ParameterListLength);
+            break;
+        }
+
+        // Copy SRB buffer if needed
+        switch(RequestType) {
+        case WnbdReqTypeWrite:
+        case WnbdReqTypePersistResIn:
```
petrutlucian94, line 222: TODO: no need to do this for PR IN commands.

stefan-chivu (approved):

PR #79, opened 2022-11-07
driver: Live resize support

stefan-chivu:
This PR contains the implementation of the live resize feature, which consists of directly modifying the value of a device's block count property, then notify the system that the size has been changed in order for it to be synchronized.


libwnbd/libwnbd.cpp
```diff
@@ -235,6 +235,44 @@ DWORD WnbdRemoveEx(
...
+
+    if (!DeviceIoControl(
+        Disk->Handle,
+        IOCTL_DISK_UPDATE_PROPERTIES,
+        NULL,
+        0,
+        NULL,
+        0,
+        &BytesReturned,
+        NULL)) {
```
petrutlucian94, line 268: nit: let's increase the parameter indentation here so that it's clearly delimited from the ``if`` code block.

driver/userspace.c
```diff
@@ -460,6 +460,27 @@ WnbdDeleteConnection(PWNBD_EXTENSION DeviceExtension,
...
+    ASSERT(DeviceExtension);
+    ASSERT(ConnectionId);
+
+    PWNBD_DISK_DEVICE Device = WnbdFindDeviceByConnId(DeviceExtension, ConnectionId, TRUE);
+
+    if (!Device) {
+        WNBD_LOG_INFO("Could not find connection to resize");
+        return STATUS_OBJECT_NAME_NOT_FOUND;
+    }
+
```
petrutlucian94, line 475: nit: let's use the following info message:

```
WNBD_LOG_INFO("Resized disk: %s. "
    "Old block count %lld, new block count: %lld. "
    "Block size: %d.",
    Device->Properties.InstanceName,
    Device->Properties.BlockCount,
    BlockCount,
    Device->Properties.BlockSize)
```

libwnbd/libwnbd.cpp
```diff
@@ -235,6 +235,80 @@ DWORD WnbdRemoveEx(
     return Status;
 }
 
+DWORD WnbdSetDiskSize(PWNBD_DISK Disk, UINT64 BlockCount) {
+    DWORD ErrorCode = ERROR_SUCCESS;
+    DWORD BytesReturned = 0;
+
+    LogDebug("Resizing device %s. New block count: %d",
```
petrutlucian94, line 242: let's use an info message instead. also, we should also log the previous block count.

libwnbd/libwnbd.cpp
```diff
@@ -235,6 +235,80 @@ DWORD WnbdRemoveEx(
...
+    }
+
+    ErrorCode = WnbdIoctlSetDiskSize(
+        Disk->Handle,
+        Disk->ConnectionInfo.ConnectionId,
+        BlockCount,
+        NULL);
+
+    if (ErrorCode) {
+        LogError("Could not set disk size.");
```
petrutlucian94, line 256: we can skip this error message, we're already logging the error in ``WnbdIoctlSetDiskSize``.

libwnbd/libwnbd.cpp
```diff
@@ -235,6 +235,80 @@ DWORD WnbdRemoveEx(
...
+        return ErrorCode;
+    }
+
+    WNBD_CONNECTION_INFO ConnectionInfo = {0};
+    ErrorCode = WnbdShow(Disk->Properties.InstanceName, &ConnectionInfo);
+
+    if (ErrorCode) {
+        LogError("Failed getting device connection info. Error: %d",
```
petrutlucian94, line 264: same here

libwnbd/libwnbd.cpp
```diff
@@ -235,6 +235,80 @@ DWORD WnbdRemoveEx(
...
+                 ErrorCode);
+        return ErrorCode;
+    }
+
+    HANDLE hDevice = NULL;
+    char deviceName[] = "\\\\.\\PhysicalDriveX";
```
petrutlucian94, line 270: why not just use a string? also, this won't work if the device number has two digits.

petrutlucian94 (changes requested):

driver/userspace.c
```diff
@@ -460,6 +460,33 @@ WnbdDeleteConnection(PWNBD_EXTENSION DeviceExtension,
...
+    PWNBD_DISK_DEVICE Device = WnbdFindDeviceByConnId(DeviceExtension, ConnectionId, TRUE);
+
+    if (!Device) {
+        WNBD_LOG_ERROR("Could not find the device to resize.");
+        return STATUS_OBJECT_NAME_NOT_FOUND;
+    }
+
+    Device->Properties.BlockCount = BlockCount;
+    WNBD_LOG_INFO("Resized disk: %s. "
```
petrutlucian94, line 477: you'll have to move the log message above the assignment, otherwise you override the old value before logging it.

petrutlucian94 (approved):

PR #84, opened 2022-12-05
driver: Added support for SCSIOP_REPORT_LUNS

stefan-chivu:
Implemented support for the REPORT_LUNS SCSI operation (A0).


driver/scsi_operation.c
```diff
@@ -547,6 +547,38 @@ WnbdModeSense(_In_ PWNBD_DISK_DEVICE Device,
...
+        SrbStatus = SRB_STATUS_INTERNAL_ERROR;
+        goto Exit;
+    }
+
+    RtlZeroMemory(DataBuffer, DataTransferLength);
+
+    PLUN_LIST LunList;
+    LunList = DataBuffer;
+    LunList->LunListLength[3] = 0x8;
+    RtlCopyMemory(&LunList->Lun[0][0], &(Device->Bus), 1);
```
petrutlucian94, line 569: I don't think the bus number should be included here.

driver/scsi_operation.c
```diff
@@ -547,6 +547,38 @@ WnbdModeSense(_In_ PWNBD_DISK_DEVICE Device,
...
+    RtlCopyMemory(&LunList->Lun[0][1], &(Device->Lun), 1);
+
+    WNBD_LOG_INFO("Device bus: %u, Device lun: %u",
+        Device->Bus, Device->Lun);
+
+    SrbSetDataTransferLength(Srb, sizeof(PLUN_LIST));
```
petrutlucian94, line 575: The data transfer length is incorrect, it should be sizeof(LUN_LIST) plus the size of the list entry.

petrutlucian94 (changes requested):

PR #88, opened 2022-12-23
tests: Added more libwnbd tests

stefan-chivu:
Added tests for testing the following:

- Live resize
- VPD 83h NAA Identifier
- Custom device serial number
- Write w/ read-only disk
- I/O stats

### DIFF
--- a/tests/libwnbd_tests/mock_wnbd_daemon.cc
+++ b/tests/libwnbd_tests/mock_wnbd_daemon.cc
@@ -42,6 +42,13 @@ void MockWnbdDaemon::Start()
         WnbdProps.Flags.FlushSupported = 1;
     }
 
+    if (UseCustomNaaIdentifier) {
+        WnbdProps.Flags.NaaIdSpecified = 1;
+        WnbdProps.NaaIdentifier.data[0] = 0x60;
+        for (int i = 1; i < sizeof(WnbdProps.NaaIdentifier.data); i++)
+            WnbdProps.NaaIdentifier.data[i] = (BYTE)rand();
+    }
+
     DWORD err = WnbdCreate(
         &WnbdProps, (const PWNBD_INTERFACE) &MockWnbdInterface,
         this, &WnbdDisk);
@@ -250,4 +257,8 @@ void MockWnbdDaemon::SendIoResponse(
 
     ASSERT_FALSE(err) << "unable to send wnbd response, error: "
                       << GetLastError();
+}
+
+PWNBD_DISK MockWnbdDaemon::GetDisk() {
+    return WnbdDisk;
 }
--- a/tests/libwnbd_tests/mock_wnbd_daemon.cc
+++ b/tests/libwnbd_tests/mock_wnbd_daemon.cc
@@ -49,6 +49,10 @@ void MockWnbdDaemon::Start()
             WnbdProps.NaaIdentifier.data[i] = (BYTE)rand();
     }
 
+    if (UseCustomDeviceSerial)
+        strcpy(WnbdProps.SerialNumber,(std::to_string(rand()) + "-"
+               + std::to_string(rand())).c_str());
+
     DWORD err = WnbdCreate(
         &WnbdProps, (const PWNBD_INTERFACE) &MockWnbdInterface,
         this, &WnbdDisk);
--- a/tests/libwnbd_tests/test.cpp
+++ b/tests/libwnbd_tests/test.cpp
@@ -781,3 +781,145 @@ void TestDeviceSerial() {
 TEST(TestDeviceSerial, TestSetDeviceSerial) {
     TestDeviceSerial();
 }
+
+void TestIoStats() {
+    auto InstanceName = GetNewInstanceName();
+
+    MockWnbdDaemon WnbdDaemon(
+        InstanceName,
+        DefaultBlockCount,
+        DefaultBlockSize,
+        false, // writable
+        true
+    );
+    WnbdDaemon.Start();
+
+    WNBD_CONNECTION_INFO ConnectionInfo = { 0 };
+    NTSTATUS Status = WnbdShow(InstanceName.c_str(), &ConnectionInfo);
+    ASSERT_FALSE(Status) << "couldn't retrieve WNBD disk info";
+
+    PWNBD_DISK WnbdDisk = WnbdDaemon.GetDisk();
+
+    std::string DiskPath = GetDiskPath(InstanceName);
+
+    WNBD_USR_STATS UserspaceStats = { 0 };
+    WNBD_DRV_STATS DriverStats = { 0 };
+
+    WnbdGetUserspaceStats(WnbdDisk, &UserspaceStats);
+    WnbdGetDriverStats(InstanceName.c_str(), &DriverStats);
+
+    // No requests have been sent, write counters should be 0
+    ASSERT_EQ(UserspaceStats.TotalWrittenBlocks, 0);
+    ASSERT_EQ(UserspaceStats.WriteErrors, 0);
+
+    HANDLE DiskHandle = CreateFileA(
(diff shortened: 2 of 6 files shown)


tests/libwnbd_tests/test.cpp
```diff
@@ -463,3 +463,39 @@ TEST(TestRead, Read2PBDisk) {
         (2LL << 50) / DefaultBlockSize,
         DefaultBlockSize);
 }
+
+void TestLiveResize(
+    uint64_t BlockCount = DefaultBlockCount,
+    uint32_t BlockSize = DefaultBlockSize,
```
petrutlucian94, line 469: nit: I think we can drop the block size parameter

tests/libwnbd_tests/test.cpp
```diff
@@ -463,3 +463,39 @@ TEST(TestRead, Read2PBDisk) {
...
+    WNBD_CONNECTION_INFO ConnectionInfo = { 0 };
+    NTSTATUS Status = WnbdShow(InstanceName.c_str(), &ConnectionInfo);
+    ASSERT_FALSE(Status) << "couldn't retrieve WNBD disk info";
+
+    PWNBD_DISK WnbdDisk = WnbdDaemon.GetDisk();
+
+    ASSERT_EQ(BlockCount, WnbdDisk->Properties.BlockCount);
+
+    ASSERT_EQ(WnbdSetDiskSize(WnbdDisk, NewBlockCount), ERROR_SUCCESS);
+    ASSERT_EQ(NewBlockCount, WnbdDisk->Properties.BlockCount);
```
petrutlucian94, line 492: We could add some assertions:

* when extending the disk, we should be able to write passed the old limit
* when shrinking the disk, we should get an error when trying to write passed the new limit

tests/libwnbd_tests/mock_wnbd_daemon.cc
```diff
@@ -42,6 +42,16 @@ void MockWnbdDaemon::Start()
         WnbdProps.Flags.FlushSupported = 1;
     }
 
+    if (InstanceName.find("TestSetNaaId") != std::string::npos) {
```
petrutlucian94, line 45: Let's avoid setting flags based on the instance name. ``SetNaaIdentifier`` is unused, you could either go with that or add another field to the constructor.

tests/libwnbd_tests/test.cpp
```diff
@@ -463,3 +521,373 @@ TEST(TestRead, Read2PBDisk) {
...
+    ASSERT_EQ(
+        sizeof(WnbdDisk->Properties.NaaIdentifier.data),
+        (unsigned int)StorageIdentifier->IdentifierSize
+    );
+    for (int i = 0; i < sizeof(WnbdDisk->Properties.NaaIdentifier.data); i++) {
+        ASSERT_EQ(
+            WnbdDisk->Properties.NaaIdentifier.data[i],
+            StorageIdentifier->Identifier[i])
+            << "VPD 83h identifiers don't match: [ 0x"
+            << byte_array_to_hex(WnbdDisk->Properties.NaaIdentifier.data, 16)
```
petrutlucian94, line 688: nit: let's use the same naming convention throughout the code.

tests/libwnbd_tests/utils.cpp
```diff
@@ -178,3 +178,11 @@ std::string GetEnv(std::string Name) {
     free(ValBuff);
     return ValStr;
 }
+
+std::string byte_array_to_hex(BYTE* arr, int length) {
+    std::stringstream ss;
+    for (int i = 0; i < length; ++i)
+        ss << std::hex << (int)arr[i];
```
petrutlucian94, line 185: nit: let's add a whitespace in between to improve readability.

tests/libwnbd_tests/utils.h
```diff
@@ -27,3 +27,7 @@ void SetDiskWritable(HANDLE DiskHandle);
 // missing.
 // Raises a runtime error upon failure.
 std::string GetEnv(std::string Name);
+
+// Returns a string containing the hex values of the byte array
+// received as parameter.
+std::string byte_array_to_hex(BYTE* arr, int length);
```
petrutlucian94, line 33: nit: let's add a newline at the end of the file.

tests/libwnbd_tests/test.cpp
```diff
@@ -463,3 +521,373 @@ TEST(TestRead, Read2PBDisk) {
...
+    WNBD_CONNECTION_INFO ConnectionInfo = { 0 };
+    NTSTATUS Status = WnbdShow(InstanceName.c_str(), &ConnectionInfo);
+    ASSERT_FALSE(Status) << "couldn't retrieve WNBD disk info";
+
+    PWNBD_DISK WnbdDisk = WnbdDaemon.GetDisk();
+
+    std::string DiskPath = GetDiskPath(InstanceName);
+
+    WNBD_USR_STATS UserspaceStats = { 0 };
+    WNBD_USR_STATS EmptyUserspaceStats = { 0 };
```
petrutlucian94, line 805: ``EmptyUserspaceStats`` is unused.

tests/libwnbd_tests/test.cpp
```diff
@@ -463,3 +521,373 @@ TEST(TestRead, Read2PBDisk) {
...
+
+    WNBD_IO_REQUEST ExpWnbdRequest = { 0 };
+    ExpWnbdRequest.RequestType = WnbdReqTypeWrite;
+    ExpWnbdRequest.Cmd.Write.BlockAddress = 0;
+    ExpWnbdRequest.Cmd.Write.BlockCount = 1;
+    // We expect to be the first ones writing to this disk.
+    ASSERT_FALSE(WnbdDaemon.ReqLog.HasEntry(ExpWnbdRequest));
+
+    DWORD BytesWritten = 0;
+    for (int i = 0; i < 1; i++) {
```
petrutlucian94, line 843: A for loop with a single iteration looks odd.

tests/libwnbd_tests/test.cpp
```diff
@@ -463,3 +521,403 @@ TEST(TestRead, Read2PBDisk) {
...
+
+    // Clear read buffer
+    memset(ReadBuffer.get(), 0, ReadBufferSize);
+    // 1 block
+    ASSERT_TRUE(ReadFile(
+        DiskHandle, ReadBuffer.get(),
+        DefaultBlockSize, &BytesRead, NULL)) << "Read failed: " << GetLastError() << std::endl;
+    ASSERT_EQ(DefaultBlockSize, BytesRead);
+
+    ASSERT_GE(UserspaceStats.TotalReadBlocks, 1);
```
petrutlucian94, line 874: I'm a bit concerned that this test could be quite flaky. The IO counters get incremented *after* the reply is submitted, so we might get outdated stats. As a workaround, we're repeating the same request, but even so we'd have to exceed the number of IO workers and it can still confuse subsequent assertions.

Probably a better approach would be to have something similar to Gomega's  ``Eventually``: https://onsi.github.io/gomega/#eventually, maybe with a 1s timeout.

tests/libwnbd_tests/test.cpp
```diff
@@ -463,3 +521,403 @@ TEST(TestRead, Read2PBDisk) {
...
+
+    // Flush twice so that there is enough time for the
+    // InvalidRequests counter to get updated
+    for (int i = 0; i < 2; i++)
+        ASSERT_FALSE(FlushFileBuffers(DiskHandle));
+
+    WnbdGetUserspaceStats(WnbdDisk, &UserspaceStats);
+
+    ASSERT_GE(UserspaceStats.InvalidRequests, 1);
+    // WIP
```
petrutlucian94, line 918: Is the WIP comment a leftover?

petrutlucian94 (approved):

PR #90, opened 2023-01-11
driver: fix scsi addressing

petrutlucian94:
This PR fixes the following SCSI address issues:

* avoid leaking addresses
* use the device list spinlock when accessing the address bitmap
* specify the right bus when notifying storport about bus changes
* use the right size when allocating the bitmap
* properly handle the addresses without making assumptions about the maximum number of luns and targets


stefan-chivu (approved):
lgtm

PR #95, opened 2023-01-13
wip: LUN level addressing

stefan-chivu:


driver/scsi_operation.c
```diff
@@ -548,25 +548,60 @@ WnbdModeSense(_In_ PWNBD_DISK_DEVICE Device,
...
-    // the LUN address is expected to be zero
     RtlZeroMemory(DataBuffer, DataTransferLength);
+    PLUN_LIST LunList = (PLUN_LIST) DataBuffer;
+    
+    ULONG LunCount = RtlNumberOfSetBits(&DeviceExtension->ScsiBitMapHeader);
+    ULONG LunListLength = 8 * LunCount;
+    KIRQL Irql = { 0 };
+    ULONG ScsiBitNumber = 0;
+
+    KeAcquireSpinLock(&DeviceExtension->DeviceListLock, &Irql);
```
petrutlucian94, line 582: ``KeAcquireSpinLock`` must be moved above RtlNumberOfSetBits.

driver/scsi_operation.c
```diff
@@ -548,25 +548,60 @@ WnbdModeSense(_In_ PWNBD_DISK_DEVICE Device,
...
     ASSERT(Srb);
 
+    UCHAR SrbStatus = SRB_STATUS_SUCCESS;
     PVOID DataBuffer = SrbGetDataBuffer(Srb);
     ULONG DataTransferLength = SrbGetDataTransferLength(Srb);
 
     if (NULL == DataBuffer) {
-        return SRB_STATUS_INTERNAL_ERROR;
+        SrbStatus = SRB_STATUS_INTERNAL_ERROR;
+        goto Exit;
```
petrutlucian94, line 564: We don't perform any cleanup in ``Exit``, why not just return directly?

driver/scsi_operation.c
```diff
@@ -548,25 +548,60 @@ WnbdModeSense(_In_ PWNBD_DISK_DEVICE Device,
...
+        ScsiBitNumber = RtlFindSetBits(
+            &DeviceExtension->ScsiBitMapHeader,
+            1,
+            ScsiBitNumber + 1);
+        if (ScsiBitNumber != 0xFFFFFFFF) {
+            LunList->Lun[i][0] = (UCHAR)(
+                            ScsiBitNumber /
+                            WNBD_MAX_LUNS_PER_TARGET /
+                            WNBD_MAX_TARGETS_PER_BUS);
+            LunList->Lun[i][1] = (UCHAR)ScsiBitNumber;
```
petrutlucian94, line 593: We're only checking it the buffer can hold a ``LUN_LIST`` structure but not the actual LUN entries.

driver/userspace.c
```diff
@@ -324,17 +323,18 @@ WnbdCreateConnection(PWNBD_EXTENSION DeviceExtension,
         return STATUS_SHUTDOWN_IN_PROGRESS;
     }
 
+    PWNBD_DISK_DEVICE Device = (PWNBD_DISK_DEVICE) Malloc(sizeof(WNBD_DISK_DEVICE));
```
petrutlucian94, line 326: no need to move this allocation here, you can just set Device to NULL so that you may use it in the ``Exit`` routine.

driver/userspace.c
```diff
@@ -469,24 +469,32 @@ WnbdCreateConnection(PWNBD_EXTENSION DeviceExtension,
     return Status;
 
 Exit:
-    if (ScsiBitNumber != 0xFFFFFFFF) {
-        KeAcquireSpinLock(&DeviceExtension->DeviceListLock, &Irql);
-        RtlClearBits(&ScsiBitMapHeader, ScsiBitNumber, 1);
-        KeReleaseSpinLock(&DeviceExtension->DeviceListLock, Irql);
-    }
+    // if monitor thread started, it will handle the clean-up
+    // otherwise, clean-up here before returning
```
petrutlucian94, line 473: nit: s/clean-up/cleanup

petrutlucian94 (changes requested):

driver/scsi_operation.c
stefan-chivu: I've added the Exit label just for logging the exit code of the function, but I guess that could be done after the function call within the caller.

driver/scsi_operation.c
```diff
@@ -548,25 +548,60 @@ WnbdModeSense(_In_ PWNBD_DISK_DEVICE Device,
...
+            "LUN list of size: %d", DataTransferLength, sizeof(LUN_LIST));
+        SrbStatus = SRB_STATUS_BAD_SRB_BLOCK_LENGTH;
+        goto Exit;
     }
 
-    // the LUN address is expected to be zero
     RtlZeroMemory(DataBuffer, DataTransferLength);
+    PLUN_LIST LunList = (PLUN_LIST) DataBuffer;
+    
+    ULONG LunCount = RtlNumberOfSetBits(&DeviceExtension->ScsiBitMapHeader);
```
petrutlucian94, line 577: Aren't we suppose to retrieve just the luns that are associated with a given target? This will go over *all* luns.

petrutlucian94:
Closing this for now for the following reasons:

* using multiple buses is an easier way of extending the adapter capacity and contrary to what we expected, it performed the same. We were expecting lun discovery to be more efficient when reporting all target luns through a single SCSI command as opposed to having Storport iterate over each bus/target.
* the PR doesn't currently handle the "select report" field
* for some reason, when having 2 targets of 255 luns, the last lun was unavailable
* as per the SCSI spec, we may have to report a virtual lun 0 even if there's no disk attached (not handled at the moment)

PR #100, opened 2023-01-25
tests: Persistent Reservation tests

stefan-chivu:
Added persistent reserve in/out tests. These tests send a passthrough SCSI command which gets processed by the mock handler in order to verify that the persistent reservation params are transmitted correctly.


tests/libwnbd_tests/scsi_helper.h
```diff
@@ -0,0 +1,17 @@
+#include "pch.h"
+#include <wnbd.h>
+
+class ScsiPassthroughHelper
+{
+private:
+	PWNBD_DISK WnbdDisk;
```
petrutlucian94, line 7: WnbdDisk seems to be unused. That being considered, not sure if we really need a class.

tests/libwnbd_tests/scsi_helper.cpp
```diff
@@ -0,0 +1,52 @@
+#include "scsi_helper.h"
+
+void ScsiPassthroughHelper::PersistentReserveIn() {
+	UCHAR cdb[16];
```
petrutlucian94, line 4: It's true that wnbd doesn't parse scsi per rsv commands but we should probably ensure that it gets transmitted properly.

tests/libwnbd_tests/test_disk_actions.cpp
```diff
@@ -457,3 +458,164 @@ TEST(TestUnmap, TestHardUnmapFallback) {
...
+
+    WnbdDaemon.Start();
+    WNBD_CONNECTION_INFO ConnectionInfo = { 0 };
+    ASSERT_FALSE(WnbdShow(InstanceName.c_str(), &ConnectionInfo))
+        << "couldn't retrieve WNBD disk info";
+
+    std::string DiskPath = GetDiskPath(InstanceName);
+    PWNBD_DISK WnbdDisk = WnbdDaemon.GetDisk();
+
+    UCHAR cdb[16];
```
petrutlucian94, line 564: ``cdb`` and ``sense`` are unused

tests/libwnbd_tests/mock_wnbd_daemon.cc
```diff
@@ -234,6 +237,89 @@ void MockWnbdDaemon::Unmap(
...
+    ASSERT_EQ(ServiceAction, RESERVATION_ACTION_CLEAR);
+    WnbdReq.Cmd.PersistResOut.ParameterListLength = ParameterListLength;
+    WnbdReq.Cmd.PersistResOut.Scope = Scope;
+    ASSERT_EQ(Scope, RESERVATION_SCOPE_LU);
+    WnbdReq.Cmd.PersistResOut.Type = Type;
+    ASSERT_EQ(Type, RESERVATION_TYPE_EXCLUSIVE);
+
+    handler->ReqLog.AddEntry(WnbdReq);
+
+    switch (ServiceAction)
```
petrutlucian94, line 301: It doesn't make much sense to have this switch statement here while having ``ASSERT_EQ(ServiceAction, RESERVATION_ACTION_CLEAR);`` a few lines above.

tests/libwnbd_tests/mock_wnbd_daemon.cc
```diff
@@ -234,6 +237,89 @@ void MockWnbdDaemon::Unmap(
...
+{
+    MockWnbdDaemon* handler = nullptr;
+    ASSERT_FALSE(WnbdGetUserContext(Disk, (PVOID*)&handler));
+
+    WnbdRequestType RequestType = WnbdReqTypePersistResIn;
+    WNBD_IO_REQUEST WnbdReq = { 0 };
+    WnbdReq.RequestType = RequestType;
+    WnbdReq.RequestHandle = RequestHandle;
+    WnbdReq.Cmd.PersistResIn.ServiceAction = ServiceAction;
+    ASSERT_EQ(ServiceAction, RESERVATION_ACTION_READ_RESERVATIONS);
```
petrutlucian94, line 253: Windows will try to read the PR keys, leading to an assertion failure.

tests/libwnbd_tests/test_disk_actions.cpp
```diff
@@ -457,3 +458,164 @@ TEST(TestUnmap, TestHardUnmapFallback) {
...
+        NULL);
+
+    EXPECT_NE(Result, 0) << "Error sending command: " << GetLastError();
+    ASSERT_EQ(sptd.ScsiStatus, 0) << "Command returned with error status: " << std::hex << sptd.ScsiStatus;
+    CloseHandle(hDevice);
+
+    //// TODO: find out why ReqLog.HasEntry returns false for the request we send
+    /*WNBD_IO_REQUEST ExpWnbdRequest = { 0 };
+    ExpWnbdRequest.RequestType = WnbdReqTypePersistResOut;
+    ExpWnbdRequest.Cmd.PersistResOut.ServiceAction = RESERVATION_ACTION_CLEAR;
```
petrutlucian94, line 615: This check fails because we aren't filling in all the expected command fields, we'll need this:

ExpWnbdRequest.Cmd.PersistResOut.Type = RESERVATION_TYPE_EXCLUSIVE;

petrutlucian94 (changes requested):

tests/libwnbd_tests/test_disk_actions.cpp
```diff
@@ -457,3 +458,154 @@ TEST(TestUnmap, TestHardUnmapFallback) {
...
+    auto InstanceName = GetNewInstanceName();
+
+    MockWnbdDaemon WnbdDaemon(
+        InstanceName,
+        DefaultBlockCount,
+        DefaultBlockSize,
+        false,
+        false,
+        false,
+        false,
```
petrutlucian94, line 472: MockWnbdDaemon arguments are getting out of hand, we'll have to merge the custom flags in a single argument. We'll do it in a separate patch.

tests/libwnbd_tests/mock_wnbd_daemon.cc
```diff
@@ -234,6 +237,85 @@ void MockWnbdDaemon::Unmap(
...
+        WnbdSetSense(
+            &handler->MockStatus,
+            SCSI_SENSE_ILLEGAL_REQUEST,
+            SCSI_ADSENSE_ILLEGAL_COMMAND);
+        break;
+    }
+
+    handler->SendIoResponse(
+        RequestHandle, RequestType,
+        handler->MockStatus, NULL, 0);
```
petrutlucian94, line 271: So we have a "persistent reservation in" command that doesn't actually return any data. It's ok for now, but we'll eventually have to ensure that persistent reservation data buffers get transferred accordingly.

PR #106, opened 2023-02-07
driver: drop firmware functions

petrutlucian94:
Some of the driver certification tests attempt to do firmware upgrades, which doesn't really make sense for our virtual adapters.

At some point, we added some fake firmware functions just to get the MS certification tests to pass, yet they currently fail anyway.

We'll go ahead and just remove the unnecessary firmware functions.


petrutlucian94:
This does actually help with one of the HLK tests. We'll keep it around, at least until we get the official confirmation that firmware update tests can be skipped for virtual devices.

PR #110, opened 2023-02-15
tests: Drop MockWnbdDaemon excessive constructor params

stefan-chivu:
The number of constructor arguments for MockWnbdDaemon was getting
out of hand, with some arguments only being necessary for one test.

With this commit, the constructor will only receive a pointer to a
WNBD_PROPERTIES structure that will contain all the properties set
accordingly for each test.

In every test, a default WNBD_PROPERTIES structure will be generated
using the GetNewWnbdProps function from utils.cpp, which receives a
reference to a WNBD_PROPERTIES structure, and for any custom
requirements, the values can be modified before calling the Start 
method of the MockWnbdDaemon class.


petrutlucian94 (approved):

petrutlucian94:
Looks good, unrelated CI failure.

PR #122, opened 2023-05-16
Move nbd to userspace

petrutlucian94:
Right now, WNBD provides two IO transmission channels:

* nbd - initial implementation
* DeviceIoControl based API - used by rbd-wnbd

rbd-wnbd no longer uses NBD, however we'll continue to offer the NBD client functionality for the following reasons:

* one of the very few Windows NBD clients out there, pretty well received by the community
* allows connecting to a variety of NBD servers, including qemu-nbd
* useful for testing purposes

We'll move the NBD client to userspace (libwnbd.dll), leveraging the DeviceIoControl API for driver communication. Main benefits:

* simplifies the driver and allows us to drop the ksocket dependency
* 20% perf improvement (WSK is quite slow, especially on older Windows versions)
* reduces the risk of crashes or security issues
* easier to maintain

The only potential downside to be aware of is that the ``wnbd-client map`` command becomes blocking. We might add an option to run the daemon in a separate process, however it would be killed when the session ends (i.e. winrm or ssh session). To avoid this, Ceph daemons are spawned by a centralized Windows service, however we aren't going to add such a service for NBD, at least for now. Other workarounds include NSSM or spawning the process using the ``Win32_Process`` WMI class.

While at it, we're bumping the c++ version to c++20, which is supported by VS 2019. Among other things, it simplifies structure
initialization.


wnbd-client/cmd.cpp
```diff
@@ -98,7 +109,7 @@ DWORD CmdMap(
     }
     if (PortNumber > 65353) {
         cerr << "Invalid NBD server port number: " << PortNumber << endl;
-        return ERROR_INVALID_PARAMETER;
+    return ERROR_INVALID_PARAMETER;
```
petrutlucian94, line 112: TODO: unrelated change.

libwnbd/nbd_daemon.cpp
```diff
@@ -0,0 +1,569 @@
...
+    return Retval;
+}
+
+DWORD NbdDaemon::Shutdown(bool HardRemove)
+{
+    std::unique_lock<std::mutex> Lock(ShutdownLock);
+
+    if (!Terminated) {
+        TerminateInProgress = true;
+        DWORD Err = DisconnectNbd();
```
petrutlucian94, line 237: TODO: we should remove the NBD connection after the WNBD one, allowing pending IO to finish.

libwnbd/nbd_daemon.cpp
```diff
@@ -0,0 +1,569 @@
...
+        Handler->Shutdown(true);
+    }
+}
+
+void NbdDaemon::NbdReplyWorker()
+{
+    // For performance reasons, we're reusing the overlapped structure
+    // when submitting IO replies to WNBD.
+    OVERLAPPED Overlapped = { 0 };
+    HANDLE OverlappedEvent = CreateEventA(0, TRUE, TRUE, NULL);
```
petrutlucian94, line 448: TODO: cleanup the event before exiting.

libwnbd/nbd_protocol.h
```diff
@@ -79,76 +88,61 @@ typedef struct _NBD_HANDSHAKE_RPL {
 __pragma(pack(pop))
 #pragma warning(default:4200)
 
-#define NBD_OPT_EXPORT_NAME	 1
-#define NBD_OPT_GO		     7
+#define NBD_OPT_EXPORT_NAME  1
```
petrutlucian94, line 91: nit: should preserve the indentation

PR #126, opened 2023-06-06
Detect and remove stale connections

petrutlucian94:
Storport resets the lun after hitting request timeouts. However,
it never actually removes the disk. Having a stale disk can be
troublesome, leading to an unresponsive host in certain situations
(e.g. cache deadlocks, hanging persistent reservation
requests, etc).

For this reason, we'll detect stale connections and disconnect
the disk. This feature along with the timeouts are configurable.
By default, we'll consider a connection to be stale if at least one
request older than 15s got aborted and if no IO reply was received
in the last minute.

At the same time, we'll include the following timestamps in the
``wnbd-client.exe stats`` output:

* last received request
* last submitted request
* last received reply

### DIFF
--- a/driver/options.h
+++ b/driver/options.h
@@ -41,6 +41,9 @@ typedef enum {
     OptDbgPrintEnabled,
     OptMaxIOReqPerAdapter,
     OptMaxIOReqPerLun,
+    OptRemoveStaleConnections,
+    OptStaleReqTimeoutMs,
+    OptStaleConnTimeoutMs,
 } WNBD_OPT_KEY;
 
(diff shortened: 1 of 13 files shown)


driver/options.h
```diff
@@ -41,6 +41,9 @@ typedef enum {
     OptDbgPrintEnabled,
     OptMaxIOReqPerAdapter,
     OptMaxIOReqPerLun,
+    OptRemoveStaleConnections,
```
petrutlucian94, line 44: TODO: document the new settings.

